PR #87, opened 2018-04-16
Add handlers and formatters that might be useful

juga0:
Though not modifying the default logger and handler

### DIFF
--- a/sbws/config.log.default.ini
+++ b/sbws/config.log.default.ini
@@ -2,10 +2,10 @@
 keys = root,sbws
 
 [handlers]
-keys = sbwsdefault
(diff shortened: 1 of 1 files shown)


pastly:
Aha! I didn't click the close button on this 3 days ago. I think it got closed because my issue68_logging branch was deleted, your issue68_logging branch was deleted, or both. This likely happened when I flipped the open source switch.

I can't reopen this, and I think it's because juga0:issue68_logging doesn't exist. I would reopen it if I could.

juga0:
Right, when making public this repo, i had to fork it again, losing the branches. I've pushed my local branches now.

PR #112, opened 2018-04-25
 Include software and sofware_version in v3bw files

### LINKED ISSUE #96: Include software and sofware_version in generate
So they would implement the new spec.

juga0:
Fixes #96.
I thought it would be better to move all that has to do with the generation of the v3bw files to ``lib/v3bwfile`` module and leave ``generate.py`` basically with only the ``main``. This way we can start to separate better what is obtaining measurements and what is generating the files for the bwauths, as suggested by arm in IRC.
Next step would be to test generate complete file including header.
For that i would remove what has to do with header in ``core/test_generate.py`` (for instance, https://github.com/pastly/simple-bw-scanner/blob/master/tests/core/test_generate.py#L101) and use the header class.
It can be done in that file, or move V3BWLine to ``lib/test_v3bwfile.py``.
@pastly what do you think?

### DIFF
--- a/sbws/core/generate.py
+++ b/sbws/core/generate.py
@@ -1,5 +1,6 @@
 from sbws import version
 from sbws.globals import (fail_hard, is_initted, time_now)
+from sbws.lib.v3bwfile import V3BwHeader
 from sbws.lib.resultdump import ResultSuccess
 from sbws.lib.resultdump import load_recent_results_in_datadir
 from sbws.lib.resultdump import group_results_by_relay
@@ -120,10 +121,10 @@ def main(args, conf):
     data_lines = [result_data_to_v3bw_line(data, fp) for fp in data]
     data_lines = sorted(data_lines, key=lambda d: d.bw, reverse=True)
     data_lines = scale_lines(args, data_lines)
+    header = V3BwHeader()
     log_stats(data_lines)
     log.info('Writing v3bw file to %s', args.output)
     with open(args.output, 'wt') as fd:
-        fd.write('{}\n'.format(int(time_now())))
-        fd.write('version={}\n'.format(version))
+        fd.write(header)
         for line in data_lines:
             fd.write('{}\n'.format(str(line)))
--- a/sbws/core/generate.py
+++ b/sbws/core/generate.py
@@ -125,6 +125,6 @@ def main(args, conf):
     log_stats(data_lines)
     log.info('Writing v3bw file to %s', args.output)
     with open(args.output, 'wt') as fd:
-        fd.write(header)
+        fd.write(str(header))
         for line in data_lines:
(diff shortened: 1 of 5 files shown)


juga0:
Ops, need to fix tests, will do that...

pastly:
It looks like the issue is straight forward.There's new lines in the v3bw file and `test_generate.py` didn't expect them to be there.

juga0:
Right, though to solve them would be nice to know what you think about previous commit

pastly:
https://trac.torproject.org/projects/tor/ticket/3723#comment:8 suggests the proposed header format is supposed to be a single line, but it looks like this PR makes a multi-line header. Is that intentional?

juga0:
Yes, i asked @teor2345 the same question: https://github.com/torproject/torspec/pull/4#issuecomment-383918069
Though i might still have not interpreted correctly the answer

teor2345:
> https://trac.torproject.org/projects/tor/ticket/3723#comment:8 suggests the proposed header format is supposed to be a single line, but it looks like this PR makes a multi-line header. Is that intentional?

No, the bandwidth-file line in the *vote* should be a single line.

The header lines in the v3bwfile must be on multiple lines for compatibility with existing Tor implementations.

juga0:
We agreed then on the format of header lines, something to change for this to get merged?.
I'd work on #119 when this get merged.

teor2345:
We need to implement one-line headers, and check that the current tor implementation can handle them.

pastly:
This is superseded by #128 and I need to not do what I did next time.

Thanks for the work you did on this. It's still getting merged in #128!

juga0:
What means, we could have all the other header lines in 1, as far as timestamp is the 1st and only have timestamp.
@pastly @teor2345 opinions?

pastly:
Either

```
1524788032
version=1.1.0 software=sbws software_version=0.1.0
node_id=$0DBFD8B8C09E579B1033EC0B0206021A5F6FA352 bw=6602 nick=exit2 rtt=20 time=1524788025
```

or

```
1524788032
version=1.1.0
software=sbws
software_version=0.1.0
node_id=$0DBFD8B8C09E579B1033EC0B0206021A5F6FA352 bw=6602 nick=exit2 rtt=20 time=1524788025
```

Seems to require 0 changes to Tor today in order for it to still technically read the file (with only warnings).

I have no strong opinion for one over the other and think we should just pick one and go with it.

I'll fixup #128 again with whatever we choose as soon as we choose it. As far as I'm concerned we could flip a coin ;)

juga0:
Sorry @teor2345 , just reading this now, this PR was already merged. And i closed #96 due that.
Reopening #96 and creating a new PR

PR #130, opened 2018-04-27
Additional headers

### LINKED ISSUE #119: Add useful information in sbws header lines
We should think about adding some more useful fields in sbws v3bw file header lines:
* when the scanner started
* the time of the earliest measurement used to generate the file
* helper exits (maybe, could have security implications)

And any other useful information.

I think the times would be useful for debugging 1.0. Other fields we could add later.

juga0:
Fixes #119 
Only adding ``earlier_result`` and ``scanner_started``.
For the 2nd, i considered different ways to store it, and thought that a file would be the simplest.
Let me know what you think.

### DIFF
--- a/sbws/config.default.ini
+++ b/sbws/config.default.ini
@@ -1,7 +1,7 @@
 [paths]
 datadir = ${sbws_home}/datadir
 v3bw_fname = ${sbws_home}/v3bw.txt
-started_filepath = {sbws_home}/started_at
+started_filepath = ${sbws_home}/started_at
 
 [destinations]
 # The path part of the URL for a destination if not specified. For example,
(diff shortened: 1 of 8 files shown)


juga0:
ops, sorry, let me fix errors first :)

juga0:
Will wait for #96 to continue with this.

juga0:
Thanks a lot for all the detailed comments! I'll work on them on when #96 is solved

juga0:
Fixed things commented that do not depend on the version.
Refactored the code to allow format header with other spec versions.
Still not to merge until we have final spec.

juga0:
Rebased with the http code.
Commits aren't very clean, if that's fine, it's ready to merge now

PR #132, opened 2018-04-28
Add what to include in a source distribution

juga0:
Useful for packaging systems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ docs/build
 .coverage
 htmlcov
 .pytest_cache
+dist
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,4 @@ include *.md
 include *.rst
 recursive-include docs *
 prune docs/build
+recursive-include tests *
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,3 +3,9 @@ include *.rst
 recursive-include docs *
 prune docs/build
 recursive-include tests *
+prune tests/testnets/*/auth?
+prune tests/testnets/*/relay?
+prune tests/testnets/*/datadir
+prune tests/testnets/*/exit?
+recursive-exclude tests/testnets/* config*.ini
+recursive-exclude tests/testnets/* *.log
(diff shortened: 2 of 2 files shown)


juga0:
Good point :)
It can be tested with ``check-manifest``, but that would add other dependency that not sure is worth.
Manually: ``python setup.py sdist`` and check the content of the tarball in ``dist/``
(https://docs.python.org/3/distutils/sourcedist.html)

pastly:
Thanks for the tip, @juga0.

I've been running integration tests, so my `tests/testnets/simple-ipv{4,6}/` directories have a bunch of crap in them. `python setup.py sdist` picks up that crap. In the last ~15m I couldn't figure out how to exclude the crap using the manifest file syntax I found [here](https://docs.python.org/3.5/distutils/sourcedist.html#specifying-the-files-to-distribute) and [here](https://docs.python.org/3.5/distutils/commandref.html#sdist-cmd). 

If you would like to see what I'm seeing, run the integration tests to generate the crap files.

```bash
# already be in your sbws virtualenv
cd tests/testnets
./run-network.sh simple-ipv4
# let it run ...
./run-network.sh simple-ipv6
# let it run ...
# now compare the contents of simple-* with what is checked into source control on GitHub
```

I can try reorganizing the placement of the generated files and directories in the integration tests if that would make including/excluding files easier for you. Just let me know.

pastly:
If it helps, here's `tree`.

https://paste.debian.net/1022549/

juga0:
Good that you tried it.
It wasn't trivial to find out how to add the rules, since ``prune`` works only for dirs and ``recursive-exclude`` for files.
Could you check it again?

juga0:
(And i'm jjust wondering was the advantage of testnets vs chutney, but that's other topic)

pastly:
Here's the full list of files in dist/sbws-0.0.0.tar.gz https://paste.debian.net/1022659/

The `__pycache__` directories are near the bottom. They only are picked up from the test directory for some reason, even though I have some `__pycache__` directories in sbws/.

I'll happily merge the rebased branch once the `__pycache__` directories are removed or I'm told it doesn't matter that they exist! I would open a new PR for it.

Thanks!

juga0:
Thanks for the feedback!
Added the rule for pycache. 
Sorry, realized there's no need for rebase, so no other PR is needed.

Issue #146, opened 2018-05-02
Use sensible tor options, and document why each one is used


teor2345:
@juga0 mentioned some tor options from bwscanner on #tor-dev. bwscanner took them from chutney and torflow. Let's think about whether we need each one:

   'LearnCircuitBuildTimeout': 0,  # Disable adaptive circuit timeouts.
    'CircuitBuildTimeout': 20,

Yes, we should have a set timeout, to avoid biasing results.

    'UseEntryGuards': 0,  # Disable UseEntryGuards to avoid PathBias warnings.

Yes, we don't need entry guards, we won't be using them.

    'UseMicroDescriptors': 0,
    'FetchUselessDescriptors': 1,

We shouldn't need these options, if Stem and sbws are coded correctly. Or, we can choose to work with them on or off, see #144.

    'FetchDirInfoEarly': 1,
    'FetchDirInfoExtraEarly': 1,

If we want to prioritise measuring new relays, we should set these options. But we should be prepared for exits to fail our circuits, because they don't have the new relay's descriptor yet.

    'SafeLogging': 0,
    'LogTimeGranularity': 1,

These are useful logging options for clients that don't care about anonymity.

teor2345:
Status update:

#147 has most of these options already.

#144 and #145 deal with the flavour options:
    'UseMicroDescriptors': 0,
    'FetchUselessDescriptors': 1,

We should use these options to measure new relays faster:
    'FetchDirInfoEarly': 1,
    'FetchDirInfoExtraEarly': 1,

We should use this option to avoid warnings:
    'UseEntryGuards': 0,  # Disable UseEntryGuards to avoid PathBias warnings.

teor2345:
Sure. That also seems like a good design choice. If sbws fetches on a client schedule, then all the relays it uses should have updated consensuses.

We only vote once an hour anyway. And it takes clients 1-2 hours to get new bandwidths after they are in a consensus.

So when fetching early, the best case delay is 3 hours, and the worst case is about 4.5 hours.
When fetching on the client schedule, the best case delay is about 4 hours, and the worse case is about 6.

You're right, the extra load isn't worth it.

PR #147, opened 2018-05-02
Make scanner launch Tor for itself

### LINKED ISSUE #77: Have sbws client launch tor itself
Stem can do it. https://stem.torproject.org/api/process.html

But I'm not sure we should. In fact, while it sounds cool, I'm leaning towards "don't do this."

Advantages:

- Most/all of the Tor options for sbws client are no longer necessary. sbws can configure Tor exactly how it wants. In fact, it can tell Tor to pick whatever it wants for a SocksPort, and sbws can use the ControlPort to figure out what was chosen)

- On the client side, the user doesn't even need to think about Tor as long as they have it installed.

- If sbws ever happened to need a custom Tor or a newer Tor than is commonly found in package managers (which I really hope is never the case), then implementing this the right way allows for the user to specify a custom path to their desired Tor

Disadvantages/unanswered questions:

- It's unclear what should happen for commands that need Tor but don't run continuously. Right now there are not any, but `sbws generate` might in the future. Should they really have to spend some time bootstrapping Tor just to determine the value of a consensus parameter? 

- What should happen if `sbws client` is running, using its Tor process, and we then run some other sbws command that requires Tor? Should another Tor process be started? Should it detect that an "sbws-owned Tor" already exists and use it?

Random thoughts about implementation details:

- Put the Tor data directory in `~/.sbws/tordata`. Could also put it in a fresh temporary directory every run, but I'm not sure that's a good idea.

- By default sbws should look for `tor` in the user's `$PATH`. But if some config option is set (say for example `tor_bin` in the section `[tor]`), it should use that instead.

- If Sbws cannot find Tor, it should only fail hard when running a command that requires it

- It's probably possible to support both "use the system Tor that's already running" and "launch our own Tor" cases, but for simplicity's sake, I think we should only support one or the other.

pastly:
Merging this closes #77

Note how most of the options in the `[tor]` section will go away. Your `[tor]` section in the config will most likely be empty now. @juga0, I highlight you because AFAIK you're the only other person running sbws occasionally.

### DIFF
--- a/tests/testnets/simple.common/01-gen-configs.sh
+++ b/tests/testnets/simple.common/01-gen-configs.sh
@@ -137,7 +137,6 @@ do
 	" >> $torrc
 done
 
-rm $auth_torrc_section
 
 # Get a random port between 2000 and 62000 while handling the fact that $RANDOM
 # doesn't go up that high
@@ -149,10 +148,8 @@ datadir = \${sbws_home}/datadir
 sbws_home = $(pwd)
 
 [tor]
-control_type = socket
-control_location = \${paths:sbws_home}/$scanner_tor_socks_proxy_nick/control_socket
-socks_host = $scanner_tor_socks_proxy_ip
-socks_port = $socksport
+extra_lines =
+$(cat $auth_torrc_section | while read LINE; do printf "    $LINE\n"; done)
 
 [scanner]
 nickname = SbwsTestnetScanner
@@ -172,6 +169,7 @@ bind_port = $sbws_server_port
 scanner1 = 9Xa9Ulp9bD5GGLuFm6XYZBtc2VhWQlJgpRRF9SpmfoujrFwdRwBizpqcSMHix6Jc
 scanner2 = gNeJoOiB7eya7QrpjtxlwSQO42eXazawJIEh5BbKJ1pZ0RFxT45Rbqv28wWyD4pk
 scanner3 = Onqr54A6xavBV5yxd4KCNPIl5mR6UdnAb21XX8t3kbEvTd28o6HQxFA2Gim8kxil
+
 
 [helpers]
 exit1 = on
@@ -197,3 +195,4 @@ server_port = $sbws_server_port
 password = Onqr54A6xavBV5yxd4KCNPIl5mR6UdnAb21XX8t3kbEvTd28o6HQxFA2Gim8kxil
 " > config.ini
 touch config.log.ini
+rm $auth_torrc_section
--- a/tests/testnets/simple.common/01-gen-configs.sh
+++ b/tests/testnets/simple.common/01-gen-configs.sh
@@ -27,7 +27,7 @@ scanner_tor_socks_proxy_ip=""
 scanner_tor_socks_proxy_nick=""
 
 echo -n '' > $auth_torrc_section
-rm -fr auth?/ relay?/ exit?/ config*.ini datadir/ *.log
+rm -fr auth?/ relay?/ exit?/ config*.ini datadir/ tor/ *.log
 for A in auth1 auth2 auth3
 do
 	mkdir -pv $A/keys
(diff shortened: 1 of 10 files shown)


tests/testnets/simple.common/01-gen-configs.sh
```diff
@@ -149,10 +148,12 @@ datadir = \${sbws_home}/datadir
 sbws_home = $(pwd)
 
 [tor]
-control_type = socket
-control_location = \${paths:sbws_home}/$scanner_tor_socks_proxy_nick/control_socket
-socks_host = $scanner_tor_socks_proxy_ip
-socks_port = $socksport
+extra_lines =
+    TestingTorNetwork 1
+    NumCPUs 1
```
teor2345: Why NumCPUs 1?
Seems like an odd option.

tests/testnets/simple.common/01-gen-configs.sh
```diff
@@ -149,10 +148,12 @@ datadir = \${sbws_home}/datadir
...
+    LogTimeGranularity 1
+    SafeLogging 0
```
teor2345: Consider UseEntryGuards 0 to disable pathbias warnings.
See #146.

tests/testnets/simple.common/01-gen-configs.sh
pastly: There's ~12 Tor processes that are run at once in these tests and I thought I would attempt to not peg all of everyone's CPU cores, if possible.

tests/testnets/simple.common/01-gen-configs.sh
pastly: @teor2345 [I don't think we should use FetchDirInfo*Early](https://github.com/pastly/simple-bw-scanner/issues/146#issuecomment-385976197)

Also, notice the inclusion of them in initial commits on this branch is undone by latter commits (ce2187e).

teor2345: I did not notice, I was using the files view.

pastly: @juga0 these lines are *in addition* to the torrc config lines you linked to. See how in [b378bf63](https://github.com/pastly/simple-bw-scanner/pull/147/commits/b378bf63e09fcbc3a060c1d387363975f89e29b9#diff-a250492b468dd26963e8a1c2343cb1e1R175) I added the ability for torrc options to be read from config.ini.

So in the test networks, all the options in `sbws/util/stem.py launch_tor()` are set, and then these `extra_lines` are added.

Does that make sense? I'm not sure I've addressed your concern.

pastly:
@juga, fcac281 addresses your suggestion on IRC to move as many torrc options as possible to a global variable.

I looked at using chutney when writing the stuff in `tests/testnets/`. Here are some of the reasons I didn't go with chutney

- It's another dependency
- It's harder to control the chutney version the user is using. It's easier to just have the test network generation scripts source-controlled right with the sbws code.
- Chutney is complex, especially in the torrc templating. It's complex because it wants to be flexible. I don't need flexible. I just want reproducible. I thought I could implement only the things I need in a couple of scripts and have those scripts be significantly easier to grok than chutney. IMHO, I succeeded. (But I also wrote them, so I'm heavily biased)

----

@juga0, feel like this is merge ready now?

juga0:
This branch was already merged.
Re. chutney, realized there isn't doc for testnets (i know i could look code...).
Is it possible right now to connect to the a running Tor (initiated by testnets) to run a unit test?

juga0:
Re Chutney dependency and versions: in bwscanner it was only being retrieved automatically by travis from git to run tests, so normal installation would not install/make use of chutney.

juga0:
>> Is it possible right now to connect to the a running Tor (initiated by testnets) to run a unit test?
> sbws scanner first tries to connect to a running Tor.
yeah, i meant unit tests, not the scanner. Rephrasing:
"Is it possible a unit test connect to a running Tor"?
> I'm working on figuring out a good way to have the unit tests launch Tor for testing. For example, each test should not necessarily have to launch Tor for itself because that would take a very long time.
Exactly, i said "connect to", and i noted that launching Tor would be slow and not reliable in: https://github.com/pastly/simple-bw-scanner/pull/174/commits/832e33902c0ecc445629a23e743c8d6760cdc435#diff-814ac08b96f7b5924db4cc9038327f11R5

I found this very easy to do with chutney

PR #163, opened 2018-05-16
Include all package data in the binary distribution

juga0:
Using MANIFEST.in would include those files in the source distribution,
but include_package_data is need to also include all of them in the
binary distribution.
The binary distrution was missing tests on others not included in
package_data.
package_data is not needed anymore.


juga0:
hmm, it works for me. Which version of pip are you using?. i've to upgrade my version to 10.0.1 to don't get it complaining about dependency_links

Issue #160, opened 2018-05-14
#16559, bwauth code needs to be smarter about failed circuits 


juga0:
@pastly this is for working on https://trac.torproject.org/projects/tor/ticket/16559.
Since we changed to use http and not helper relay (or now is optional, i dunno), could you update us on the status of this and/or point to the code that checks when a circuit fail or a download fail?.
If there're not such checks i'd create them, and if can't know so far the reason for failure, i'd try to include it so we can debug better

pastly:
https://github.com/pastly/simple-bw-scanner/blob/2727e20ed54df9d37401035b97a6a67fd4ec60a6/sbws/core/scanner.py#L131 is `measure_relay`, which does all the circuit building and measuring. See where it calls `measure_rtt_to_server` and `measure_bandwidth_to_server` and see where it returns `ResultError` types early if something goes wrong.

pastly:
From trac#16559 ...

> it will still be counted as a circuit that went through all of the nodes [...]  the code will consider failures from unstable nodes to be relevant for nodes that are perfectly stable

sbws has the same limitation.

juga0:
hmm, not sure what "relevant" means here.
Maybe we could check whether the relays have the stable flag?
After looking the code, i found the possible errors:
- unable to build circuit between relay and exit: ResultErrorCircuit
    - PathLengthException
    - is_controller_okay: have to look what this means
    - InvalidRequest, CircuitExtensionFailed, ProtocolError: this errors can tell us more about what failed?
    - is c.new_circuit able to tell us more?
- web server usable: ResultErrorStream: have to look what this means
- measure bw and rtt: ResultErrorStream
    - requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout: this errors can tell us more about what failed?
    - other possible errors with requests?

pastly:
In [this](https://github.com/pastly/simple-bw-scanner/issues/160#issuecomment-388827299) comment I'm saying that if one relay in a circuit causes or experiences a problem, all relays in the circuit will get a ResultError* given to it.

[...]

PR #174, opened 2018-05-28
Add function to obtain relay ed25519

juga0:
Not sure this is the best way to obtain it. Please review.


sbws/core/scanner.py
```diff
@@ -131,7 +132,25 @@ def measure_bandwidth_to_server(session, conf, dest, content_length):
...
+def obtain_ed25519(controller, relay):
+    """Obtain ed25519 identity of the relay. Modifies relay object.
+
+    :param stem.control.Controller controller: stem Tor _init_controller
+    :param Relay: the relay
+    """
+    if relay.identifier is None:
+        try:
+            log.debug('Getting microdescriptor to obtain ed25519 identity.')
+            mdesc = controller.get_microdescriptor(relay.fingerprint)
```
teor2345: This code only works for clients with "UseMicrodescriptors 1" (or auto), which is the default.

But many torflow clients have "UseMicrodescriptors 0", and we want to support clients that have server descriptors. So if getting the microdescriptor fails, please fall back to getting the server descriptor:

https://stem.torproject.org/api/control.html#stem.control.Controller.get_server_descriptor

teor2345: You'll want the ed25519_master_key of the RelayDescriptor:
https://stem.torproject.org/api/descriptor/server_descriptor.html#stem.descriptor.server_descriptor.RelayDescriptor

sbws/core/scanner.py
```diff
@@ -131,7 +132,25 @@ def measure_bandwidth_to_server(session, conf, dest, content_length):
...
+            if 'ed25519' in mdesc.identifiers.keys():
+                relay.identifier_type = 'ed25519'
+                relay.identifier = mdesc.identifiers['ed25519']
+        except DescriptorUnavailable as e:
+            log.debug('e: %s')
+
+
 def measure_relay(args, conf, destinations, cb, rl, relay):
+    obtain_ed25519(cb.controller, relay)
```
teor2345: Please also try to obtain the ed25519 id when the relay object is created.

sbws/core/scanner.py
```diff
@@ -152,6 +171,7 @@ def measure_relay(args, conf, destinations, cb, rl, relay):
         # TODO: Return ResultError of some sort
         return None
     exit = rng.choice(exits)
+    obtain_ed25519(cb.controller, exit)
```
teor2345: Please also try to obtain the ed25519 id when the relay object is created.

sbws/core/scanner.py
```diff
@@ -131,7 +132,25 @@ def measure_bandwidth_to_server(session, conf, dest, content_length):
...
+    :param stem.control.Controller controller: stem Tor _init_controller
+    :param Relay: the relay
+    """
+    if relay.identifier is None:
+        try:
+            log.debug('Getting microdescriptor to obtain ed25519 identity.')
+            mdesc = controller.get_microdescriptor(relay.fingerprint)
+            if 'ed25519' in mdesc.identifiers.keys():
+                relay.identifier_type = 'ed25519'
+                relay.identifier = mdesc.identifiers['ed25519']
```
teor2345: Maybe the function should strip any trailing = signs here.

sbws/core/scanner.py
juga0: I wanted to do like that, but Relay object is only created at the end, when dumping results, as part of Result object: https://github.com/pastly/simple-bw-scanner/blob/master/sbws/lib/resultdump.py#L150
The first place where 1st is obtained descriptors (not relays) is in https://github.com/pastly/simple-bw-scanner/blob/master/sbws/lib/relaylist.py#L134
Without refactoring code, 2 possible solutions:
- obtain ed25519 when descriptors are obtained, and modify descriptor object
- obtain ed25519 in Result and/or Relay object
Which one seems better?

juga0: or other ideas where to do it? (@pastly)

juga0:
Regarding the tests, this was WIP and there isn't any test for scanner.py, where i created it.
I'm not sure which is the best way right now to test this, since we aren't using chutney (testnets instead) and it would need to launch or connect to Tor (don't think i can mock it).
Unless there're other better ideas, i think i'd launch tor in conftest.py so it can be reused for other tests.

juga0:
Instead of obtaining the key in ``scanner``, now  it's obtained when creating the ``RelayList``.
I added a test that needs to start Tor. While not using ``chutney``, i don't know a better way to do this.
I'm not modifying the Bandwidth List file lines, since that part of the code will change after merging #175 and #176. I'd modify that once those PRs are merged.

juga0:
Travis fail due tor not being installed. I could add that to Travis, though i'd also add ``chutney``

pastly:
We've decided this doesn't matter for minimal viable product (1.0) so I'm removing that as a milestone.

juga0:
Rebased to master.

sbws/core/generate.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, fp, bw, nick, rtts, last_time):
         rtts = [round(r * 1000) for r in rtts]
         self.rtt = round(median(rtts))
         self.time = unixts_to_isodt_str(last_time)
+        self.ed25519 = ed25519
 
     def __str__(self):
```
pastly: The point of getting the ed25519 identity is to include it on the line, right? So if it is not None, we should include it here, right?

sbws/lib/relaylist.py
```diff
@@ -11,6 +13,18 @@
 log = logging.getLogger(__name__)
 
 
+class RelayNS(RouterStatusEntryV3):
```
pastly: I think instead of inheriting RouterStatusEntryV3, we should

- not inherit anything
- rename to Relay (since we're going to pull info from more than the network status consensus)
- init with just a fingerprint
- use the fingerprint to ask stem for relay information (like with `get_network_status` and/or `get_server_descriptor` and/or `get_microdescriptor`) store what we get
- define `@property`s for all the things we need like exit_policy and fingerprint and nickname and flags and its ed25519 certificate and ...

sbws/lib/relaylist.py
```diff
@@ -68,6 +82,40 @@ def measured(self):
         # return [r for r in relays if r.measured is not None]
         return [r for r in relays if not r.is_unmeasured]
 
+    def relay_ed25519_master_key(self, ns):
```
pastly: I think this function doesn't belong in RelayList (it does something for a single relay, not something related to the RelayList as a whole); if the relay class is implemented as I envision, this function will aide the creation of `Relay.__init__` and then disappear.

sbws/core/generate.py
juga0: yes, the point it's to print it in the bw line.
i think i put a comment somewhere that i was not adding that until #160 was merged, cause #160 was rewriting all that code.
i should rebase this to master again and print it in the bw line

sbws/lib/relaylist.py
juga0: i agree, i put it there cause is where currently the info for bw lines is coming from.

sbws/lib/relaylist.py
juga0: sounds good, maybe we should rename some "relay" vars that won't be Relay objs after these changes (not sure if/which now)

pastly:
I'm inclined to delete the tests that are partly causing the merge conflict.

juga0:
> i think i put a comment somewhere that i was not adding that until #160 was merged, cause #160 was rewriting all that code.
> i should rebase this to master again and print it in the bw line

@pastly you forgot to rebase to master first. 
I've done it in https://github.com/juga0/simple-bw-scanner/tree/obtain_ed25519_rebased_master.
I was going to cherry-pick your commits and modify as needed, but maybe you prefer to do itf?

juga0:
I tried rebasing first on top of your changes, but it was too many conflicts

juga0:
ok, i'm cherry-picking your commits, resolving conflicts and solving new thinks that broke. Will push in the next hours

pastly:
Thanks @juga0. You can delete tests that are hard to fix.

juga0:
Rebased also in this branch, cherry-picked your commits (solving merge conflicts) and fixed some tests.
The last two commits contain FIXMEs about tests not working because of Result changes.
Would you check if easy to fix for you?, since you created/modified that class.
Otherwise let me know and i fix them.
Thanks!

juga0:
I'm fixing them

juga0:
oh!, i didn't say it was ready to merge yet, i'm still fixing things...

pastly:
Crap I need to communicate better with you.

Well I'll merge those too when you're ready.

juga0:
ok, thx

juga0:
Last commit was just s/ed25519_master_key/master_key_ed25519 to unify the terms. Except for ``_from_desc('ed25519_master_key')``, as stem uses that term instead.
If you are fine with last commit, is ready to merge.
I don't see last commits here, maybe because it has been merged, but they're in my branch.

pastly:
Grabbed your docstring commit and the variable renaming commit. Thanks for all your work on this issue.

PR #185, opened 2018-06-07
Use keyserver pool to avoid single server timeout

juga0:
travis test failed because keys.gnupg.net timeout.
It should not be the case using the pool.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install: pip install tox-travis
 script: tox
 
 before_install:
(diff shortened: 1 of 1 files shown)


juga0:
dammit, still gpg version 1, anyways, now no errors and no timeout

pastly:
Thanks! Hope this makes things better. I noticed the issue too and thought about including a copy of the public key somehow. Maybe using a pool will be enough.

Issue #179, opened 2018-06-05
Keep previous v3bw.txt files 


juga0:
In https://trac.torproject.org/projects/tor/ticket/21378, the bwauths should keep the files used in their votes.

Currently sbws overwrites any previous ``v3bw.txt`` file when generating a new one.

I think it would be easier for the bwauths to keep the versions used in their votes if ``sbws`` while name the generated files with a timestamp, for instance ``20180605_1150_v3bw.txt`` (not sure which datetime name schema would be better here).

I'm not sure if the latest one should be a link, for instance ``v3bw.txt`` -> ``20180605_1150_v3bw.txt``.

Renaming oldest file according to the time they were generated would be more code complexity, so i don't think we should do it.

pastly:
I hear you're going to work on this. Yay!

I think the config should have a path to a directory to store v3bw files in. For example, remove `v3bw_fname` and replace it with

```
[paths]
v3bw_dname = ${sbws_home}/v3bw
```

Then in there I'd write files with the current date and time in the name somehow (`YYYYMMDD_HHMM_v3bw.txt` seems fine to me). Finally, I'd have a `v3bw.txt` in there that is always a symlink to the newest file.

```
/home/user/.sbws/
    v3bw/
        v3bw.txt -> 20181010_0000_v3bw.txt
        20181010_0000_v3bw.txt
        20180909_0000_v3bw.txt
```

juga0:
this is as i was doing right now:
   v3bw_dir = ${sbws_home}/v3bw 
   v3bw_fname = ${v3bw_dir}/v3bw_{}

The ``{}`` gets replaced by the timestamp.
i'm going to change ``v3bw_dir`` for `v3bw_dname`` and then would be mostly as you say except the date will be at the end.
Since we have a dir, maybe the name of the file could be just the timestamp as ``torflow``, and maybe i could format it also as ``torflow``: ``2018-06-13-05-10-00``?

juga0:
Correct myself, ``torflow`` generates the files as ``bwscan.20180603-1645``, but it does not look nice, so i'd rather format the files as ``20180909_0000``.
would be this fine for you?

pastly:
My new favorite option is `YYYYMMDD_HHMM.v3bw`

```
/home/user/.sbws/
    v3bw/
        latest.v3bw -> 20181010_0000.v3bw
        20181010_0000.v3bw
        20180909_0000.v3bw
```

[...]

PR #190, opened 2018-06-14
Issue179

juga0:


juga0:
Fixes #179

juga0:
Tests pass, ready to merge from my side

pastly:
Can you add a `latest.v3bw` symlink? After that I'll merge.

That was easier than I thought it would be!

juga0:
Oh sorry, forgot...

juga0:
and i probably should also lock the dir before writing?

pastly:
> and i probably should also lock the dir before writing?

Yeah. Good catch.

juga0:
I started to change ``test_generate.py`` to don't use ``/dev/stdout`` (so that ``DirectoryLock`` can work), but it was lot of changes/work, so i just added a condition in ``write``

juga0:
Ready to merge

pastly:
`latest.v3bw` is created in the pwd, not the `v3bw_dname`.

If `latest.v3bw` already exists, a traceback is generated instead of removing and recreating it

juga0:
ops, let me fix it

juga0:
Should be ready...

PR #199, opened 2018-06-18
Issue154

### LINKED ISSUE #154: sbws should reset relay bandwidths when their IPv4 address changes
This should be configurable by IP version.

At the moment, the defaults should be to reset on IPv4 change, but leave it alone on IPv6 change.

See trac 22308 for a similar issue with uptimes.

juga0:
Fixes #154 
@pastly  please, let me know if you would like modifications to these changes and i do them.
If you would like to modify them, it's probably better to merge this PR, and then open other PR with your changes.

### DIFF
--- a/sbws/config.default.ini
+++ b/sbws/config.default.ini
@@ -23,7 +23,13 @@ http_timeout = 10
 # CircuitBuildTimeout doesn't handle the case of a TLS connection to a relay
 # taking forever, and probably other not-yet-discovered cases.
 circuit_timeout = 10
+# Whether or not to reset the bandwidth measurements when the the relay's IPv4
+# changes
 reset_bw_ipv4_changes = on
+
+# Whether or not to reset the bandwidth measurements when the the relay's IPv6
+# changes.
+# This is currently not implemented.
 reset_bw_ipv6_changes = off
 
 [scanner]
--- a/sbws/config.default.ini
+++ b/sbws/config.default.ini
@@ -23,6 +23,8 @@ http_timeout = 10
 # CircuitBuildTimeout doesn't handle the case of a TLS connection to a relay
 # taking forever, and probably other not-yet-discovered cases.
 circuit_timeout = 10
+reset_bw_ipv4_changes = on
+reset_bw_ipv6_changes = off
 
 [scanner]
 # A human-readable string with chars in a-zA-Z0-9 to identify your scanner
--- a/sbws/lib/resultdump.py
+++ b/sbws/lib/resultdump.py
@@ -500,9 +500,10 @@ class ResultDump:
                 self.data[fp] = []
             self.data[fp].append(result)
             self.data = trim_results(self.fresh_days, self.data)
-            # we probably do not want to remove the results for a relay
-            # that has changed address when storing the results
-            # it will be already done when loading
+            # Not calling trim_results_ip_changed here to do not remove
+            # the results for a relay that has changed address.
+            # It will be called when loading the results to generate a v3bw
+            # file.
 
     def handle_result(self, result):
         ''' Call from ResultDump thread. If we are shutting down, ignores
--- a/sbws/lib/resultdump.py
+++ b/sbws/lib/resultdump.py
@@ -144,6 +144,9 @@ def load_recent_results_in_datadir(fresh_days, datadir, success_only=False,
                 results = merge_result_dicts(results, new_results)
         working_day += timedelta(days=1)
     results = trim_results(fresh_days, results)
+    # in time fresh days is possible that a relay changed ip,
+    # if that's the case, keep only the results for the last ip
+    results = trim_results_ip_changed(results, ipv4, ipv6)
     num_res = sum([len(results[fp]) for fp in results])
     if num_res == 0:
         log.warning('Results files that are valid not found. '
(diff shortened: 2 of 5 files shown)


sbws/config.default.ini
```diff
@@ -23,6 +23,8 @@ http_timeout = 10
 # CircuitBuildTimeout doesn't handle the case of a TLS connection to a relay
 # taking forever, and probably other not-yet-discovered cases.
 circuit_timeout = 10
+reset_bw_ipv4_changes = on
+reset_bw_ipv6_changes = off
```
pastly: Please leave a comment here documenting the purpose of these two options

pastly: Please add these options to `_validate_general` in `sbws/util/config.py`. Otherwise, you get this when running sbws.

```
$ sbws generate
general/reset_bw_ipv4_changes (on): Unknown key
general/reset_bw_ipv6_changes (off): Unknown key

File: sbws/lib/resultdump.py [hunk 2]
```diff
@@ -82,7 +82,44 @@ def trim_results(fresh_days, result_dict):
     return out_results
 
 
-def load_recent_results_in_datadir(fresh_days, datadir, success_only=False):
+def trim_results_ip_changed(result_dict, ipv4=True, ipv6=False):
```
pastly: Please rename ipv4/ipv6 to be more descriptive. For example, `trim_changed_ipv4` or `on_changed_ipv6`

pastly: Consider making both default to False (but leaving the ipv4 option True in config.default.ini

sbws/lib/resultdump.py
```diff
@@ -82,7 +82,44 @@ def trim_results(fresh_days, result_dict):
...
+    create a new results' dictionary without that relay's results using an
+    older IP.
+
+    :param dict result_dict: a dictionary of results
+    :param bool ipv4: whether to trim the results when a relay's IPv4 changes
+    :param bool ipv6: whether to trim the results when a relay's IPv6 changes
+    :returns: a new results dictionary
+    """
+    assert isinstance(result_dict, dict)
+    assert ipv4 is True or ipv6 is True
```
pastly: If you make both default to False, we can return early here.

sbws/lib/resultdump.py
```diff
@@ -82,7 +82,44 @@ def trim_results(fresh_days, result_dict):
...
+                # a relay that change address
+                ordered_results = sorted(results, key=lambda r: r.time)
+                latest_address = ordered_results[-1].address
+                last_ip_results = [result for result in results
+                                   if result.address == latest_address]
+                new_results_dict[fp] = last_ip_results
+            else:
+                new_results_dict[fp] = results
+    if ipv6 is True:
+        # Not implemented
```
pastly: Consider log.warning with something like "We have not implement trimming results based on a relay changing its IPv6 address at this time."

Consider also raising NotImplementedError.

sbws/lib/resultdump.py
```diff
@@ -82,7 +82,44 @@ def trim_results(fresh_days, result_dict):
...
+        pass
+    return new_results_dict
+
+
+def load_recent_results_in_datadir(fresh_days, datadir, success_only=False,
+                                   ipv4=True, ipv6=False):
```
pastly: Please give ipv4/ipv6 better names.

Consider having both default to False.

sbws/lib/resultdump.py
```diff
@@ -455,6 +495,9 @@ def store_result(self, result):
                 self.data[fp] = []
             self.data[fp].append(result)
             self.data = trim_results(self.fresh_days, self.data)
+            # we probably do not want to remove the results for a relay
+            # that has changed address when storing the results
+            # it will be already done when loading
```
pastly: Right.

Please add to this comment the fact that we are not calling `trim_results_ip_changed` because [reasons you already stated]

tests/unit/lib/test_resultdump.py
```diff
@@ -0,0 +1,31 @@
...
+                               'sbws', t=TIME1)
+RESULTSUCCESS2 = ResultSuccess(RTTS, DOWNLOADS, RELAY2, CIRC, DEST_URL,
+                               'sbws', t=TIME2)
+RESULTDICT = {FP1: [RESULTSUCCESS1, RESULTSUCCESS2]}
+
+
+def test_trim_results_ip_changed():
+    expected_results_dict = {FP1: [RESULTSUCCESS2]}
+    results_dict = trim_results_ip_changed(RESULTDICT)
+    assert expected_results_dict == results_dict
```
pastly: Thank you for writing a test!

Please also test that the function correctly doesn't do anything wrong in the case that there doesn't need to be anything trimmed.

pastly:
To add on to the review comment above ...

I actually feel kinda strongly about ipv4/ipv6 both defaulting to False in the code (and having the option for ipv4 set to True in the config.default.ini). I feel this way for two reasons:

- I think the function in question should default to doing nothing.
- In the future, if we add even more code that actually relies on the in-code default values, the sbws maintainer might want to change the defaults and get confused when he sets the ipv4 option to False in config.deafult.ini but continues to see behavior as if the ipv4 option is True.

juga0:
thanks!, i agree with al suggestions, will make those changes soon
(maybe after we finish with the bugs in 1.0 must milestone)

sbws/lib/resultdump.py
teor2345: In the public network, adding or removing an IPv6 address shouldn't change the load on a relay very much. But we will eventually add more features that use IPv6.

When IPv6 is more than a few percent of relay load, we should reset on IPv6 as well.

I added #202 so we remember to implement this feature.

sbws/config.default.ini
teor2345: Please add a comment that IPv6 trimming isn't implemented yet.
Otherwise, operators will set it to on, and be surprised when they get runtime errors.

juga0:
What you suggested should be fixed now

PR #212, opened 2018-06-25
Catch JSON decode error

### LINKED ISSUE #210: Raised TypeError in resultdump
Running ``sbws`` 09691a0fe7b3809f2cdacd7713c2b37668c6c93b with python 3.5.3 in Tor network got the following:

```
[2018-06-23 14:08:33,325] [sbws.util.filelock] [DEBUG] Releasing lock  /path/.sbws_tor_35/datadir/.lockfile
Exception in thread Thread-1:
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File " /path/simple-bw-scanner/sbws/lib/resultdump.py", line 478, in enter
    self.fresh_days, self.datadir)
  File " /path/simple-bw-scanner/sbws/lib/resultdump.py", line 106, in load_recent_results_in_datadir
    fname, success_only=success_only)
  File " /path/simple-bw-scanner/sbws/lib/resultdump.py", line 47, in load_result_file
    r = Result.from_dict(json.loads(line.strip()))
  File "/usr/lib/python3.5/json/__init__.py", line 319, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.5/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.5/json/decoder.py", line 355, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 502 (char 501)

  File " /path/.virtualenvs/simplebwscanner3/bin/sbws", line 11, in <module>
    load_entry_point('sbws', 'console_scripts', 'sbws')()
  File " /path/simple-bw-scanner/sbws/sbws.py", line 51, in main
    exit(comm['f'](*comm['a'], **comm['kw']))
  File " /path/simple-bw-scanner/sbws/core/scanner.py", line 402, in main
    run_speedtest(args, conf)
  File " /path/simple-bw-scanner/sbws/core/scanner.py", line 362, in run_speedtest
    for target in rp.best_priority():
  File " /path/simple-bw-scanner/sbws/lib/relayprioritizer.py", line 55, in best_priority
    results = rd.results_for_relay(relay)
  File " /path/simple-bw-scanner/sbws/lib/resultdump.py", line 503, in results_for_relay
    if fp not in self.data:
TypeError: argument of type 'NoneType' is not iterable
```
i've not investitgated what caused it, but it seems we should check that self.data is not None.

I think this has to be in milestone 1, even if it seems non-ending...

juga0:
Fixes #210.

I'm not adding a test for this since we need to change the tests for result.

### DIFF
--- a/sbws/lib/resultdump.py
+++ b/sbws/lib/resultdump.py
@@ -44,7 +44,11 @@ def load_result_file(fname, success_only=False):
         with open(fname, 'rt') as fd:
             for line in fd:
                 num_total += 1
-                r = Result.from_dict(json.loads(line.strip()))
+                try:
+                    r = Result.from_dict(json.loads(line.strip()))
+                except json.decoder.JSONDecodeError:
+                    log.warning('Could not decode result %s', line.strip())
+                    r = None
                 if r is None:
                     num_ignored += 1
                     continue
--- a/sbws/lib/resultdump.py
+++ b/sbws/lib/resultdump.py
@@ -497,7 +497,7 @@ class ResultDump:
         self.fresh_days = conf.getint('general', 'data_period')
         self.datadir = conf['paths']['datadir']
         self.end_event = end_event
-        self.data = None
+        self.data = {}
         self.data_lock = RLock()
         self.thread = Thread(target=self.enter)
         self.queue = Queue()


sbws/lib/resultdump.py
```diff
@@ -560,6 +564,6 @@ def results_for_relay(self, relay):
         assert isinstance(relay, Relay)
         fp = relay.fingerprint
         with self.data_lock:
-            if fp not in self.data:
+            if self.data is None or fp not in self.data:
```
pastly: There are more places where we access self.data than just here. We also access it in self.store_result().

Instead of finding all the places where we use self.data and checking if it is None, I'd prefer either

1. letting the exception happen causing a traceback and sbws to crash. The only time self.data should be None is if it was unable to get initialized with load_recent_results_in_datadir, in which case there's probably something very wrong and we shouldn't try to recover.

2. Initialize self.data to an empty dictionary in `ResultDump.__init__`.

I like option 2 more right now.

juga0:
I agree with the reasons and i changed the commit to do as in 2.

Issue #209, opened 2018-06-23
Maybe unify integration and unit conftests


juga0:
There're basically duplicated fixtures in integration/conftest and unit/conftest.
Probably it's possible to have a common conftest in tests/ with them and just use different args with other fixture if needed.

juga0:
I've done this, but not doing PR yet since need to modify/ensure rest of tests continue working

PR #213, opened 2018-06-25
Add write method back to V3BWFile clas

### LINKED ISSUE #197: Add write method back to V3BWFile class
See discussion [here](https://github.com/pastly/simple-bw-scanner/commit/873530454b07d99674f805b8652888b866cb9683#r29376957).

To solve the duplicate code due to special handling of stdout, I propose removing the ability to write to stdout out.

juga0:
Fixes #197

@pastly, with the introduced fixtures we could rewrite empty_sbws and others in the future, unless you have other suggestions.

### DIFF
--- a/sbws/core/generate.py
+++ b/sbws/core/generate.py
@@ -4,7 +4,6 @@ from sbws.lib.resultdump import load_recent_results_in_datadir
 from argparse import ArgumentDefaultsHelpFormatter
 import os
 import logging
-from sbws.util.filelock import DirectoryLock
 from sbws.util.timestamp import now_fname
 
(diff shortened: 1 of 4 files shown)


pastly:
I'm putting off thinking about what to do about duplicate code for tests. We have #209 and #216 already for stuff of that nature.

juga0:
> I'm putting off thinking about what to do about duplicate code for tests. We have #209 and #216 already for stuff of that nature.

they are indeed related and probably an probably can't separate the work
or one would depend on the other getting merge.

If it's fine for you in any case, i'd work on them, aka trying to have
better/more unit tests.

PR #215, opened 2018-06-25
Add to contributing guidelines, especially about git/github

### LINKED ISSUE #208: Add to CONTRIBUTING which git workflow is prefered and on what to follow PEP8
So far we've being using rebase workflow, but it's not specified in CONTRIBUTING if it's prefered instead of merge.
The line ``We use flake8 to enforce standard python coding style.`` should be changed to:
- on what we use PEP8 (not all of it is being followed)
- that flake8 is used to check some PEP8 errors

pastly:
Closes #208

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -26,6 +26,9 @@ Pull requests are welcome, especially if they address open issues.
 We use flake8 to check some PEP8 errors/warnings. This will be checked with
 ``tox`` and Travis.
 
+Simple Bandwidth Scanner is moving towards 100% test coverage. Please help us
+reach that goal, or at least not move us away from it.
+
 **Write good commit messages** that at least follow the spirit of
 :ref:`this example <commit-msg>`.
 
@@ -33,8 +36,6 @@ We use flake8 to check some PEP8 errors/warnings. This will be checked with
 large mega-commit. This not only makes code review easier, but it also makes
 commits that show up in ``git blame`` 10 years from now make more sense.
 
-Simple Bandwidth Scanner is moving towards 100% test coverage. Please help us
-reach that goal, or at least not move us away from it.
 
 Coding Guidelines
 =================
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -29,13 +29,28 @@ We use flake8 to check some PEP8 errors/warnings. This will be checked with
 Simple Bandwidth Scanner is moving towards 100% test coverage. Please help us
 reach that goal, or at least not move us away from it.
 
+Git and GitHub Guidelines
+=========================
+
 **Write good commit messages** that at least follow the spirit of
 :ref:`this example <commit-msg>`.
 
-**Write many small commits** each containing an atomic change instead of one
-large mega-commit. This not only makes code review easier, but it also makes
-commits that show up in ``git blame`` 10 years from now make more sense.
+Strive to **write many small commits** each containing an atomic change instead
+of one large mega-commit. This not only makes code review easier, but it also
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -29,13 +29,28 @@ We use flake8 to check some PEP8 errors/warnings. This will be checked with
...
+Git and GitHub Guidelines
+=========================
+
 **Write good commit messages** that at least follow the spirit of
 :ref:`this example <commit-msg>`.
 
-**Write many small commits** each containing an atomic change instead of one
-large mega-commit. This not only makes code review easier, but it also makes
-commits that show up in ``git blame`` 10 years from now make more sense.
+Strive to **write many small commits** each containing an atomic change instead
```
juga0: However, small commits makes not clear what is a complete feature.
In other projects, any commit should pass all tests, and if adding a new feature, it should contain also the test. Following that, commits can be actually quite big.
So far i never heard of a workflow that would prioritize ``git blame``.  If a project is mature/in production, developers try to modify code making as less changes as possible to other parts of the code (probably made by other developers).

CONTRIBUTING.rst
```diff
@@ -29,13 +29,28 @@ We use flake8 to check some PEP8 errors/warnings. This will be checked with
...
+of one large mega-commit. This not only makes code review easier, but it also
+makes commits that show up in ``git blame`` 10 years from now make more sense.
+
+**Prefer a rebase workflow instead of merge**. Incorporating PRs should be done
```
juga0: Rebase workflow makes commits cleaner, however, unless done in different branches as you commente below, it would imply "lying" for remote branches.

CONTRIBUTING.rst
```diff
@@ -29,13 +29,28 @@ We use flake8 to check some PEP8 errors/warnings. This will be checked with
...
+with fast-forward merge, if easily possible. The larger the topic branch, the
+harder this may be, so merge commits are allowed.
+
+If, while working on a topic branch, some changes are made to master that
+conflict with your work or that you need to incorporate into your work, **do
+not merge master into your topic branch**; instead, rebase your topic branch on
+top of master or cherry-pick the changes.
 
+**Do not force push lightly** unless branches are clearly labeled as ones that
```
juga0: we haven't being following this so far, personally i didn't cause it was just you and me.

CONTRIBUTING.rst
```diff
@@ -26,15 +26,16 @@ Pull requests are welcome, especially if they address open issues.
 We use flake8 to check some PEP8 errors/warnings. This will be checked with
 ``tox`` and Travis.
 
+Simple Bandwidth Scanner is moving towards 100% test coverage. Please help us
```
juga0: We're at 68% currently, and i think we were already when you wrote that.
I'd rather say that any new feature should come with the test

CONTRIBUTING.rst
```diff
@@ -40,7 +40,7 @@ commits that show up in ``git blame`` 10 years from now make more sense.
 Coding Guidelines
 =================
 
-**Strongly prefer Unix timestamps** over ``datetime`` structures and always
+**Strongly prefer Unix timestamps or datetime objects in UTC** and always
```
juga0: and that means using pytz, what has been only needed when converting from unix timestamp to datetime, but not the other way around, because there was not functionality in datetime for that.
However, working on UTC when possible, it's right.

juga0:
An example: https://libcloud.readthedocs.io/en/latest/development.html#contributing

juga0:
I think this text is fine, though could be more detailed.
Merging now, maybe when we are done with milestone 1.0, i would propose additional text to add.

PR #220, opened 2018-07-01
Check that there are exits

### LINKED ISSUE #211: Raised IndexError when choosing exits in test network
Running sbws 09691a0fe7b3809f2cdacd7713c2b37668c6c93b with python 3.5.3 in the test network got this traceback:

```
    i = self._randbelow(len(seq))
  File " /path/.virtualenvs/simplebwscanner3/lib/python3.5/random.py", line 239, in _randbelow
    r = getrandbits(k)          # 0 <= r < 2**k
  File " /path/.virtualenvs/simplebwscanner3/lib/python3.5/random.py", line 663, in getrandbits
    raise ValueError('number of bits must be greater than zero')
ValueError: number of bits must be greater than zero

During handling of the above exception, another exception occurred:

  File " /path/simple-bw-scanner/sbws/core/scanner.py", line 257, in dispatch_worker_thread
    return measure_relay(*a, **kw)
  File " /path/simple-bw-scanner/sbws/core/scanner.py", line 172, in measure_relay
    dest = destinations.next()
  File " /path/simple-bw-scanner/sbws/lib/destination.py", line 245, in next
    self._perform_usability_test()
  File " /path/simple-bw-scanner/sbws/lib/destination.py", line 193, in _perform_usability_test
    exit = self._rng.choice(exits)
  File " /path/.virtualenvs/simplebwscanner3/lib/python3.5/random.py", line 264, in choice
    raise IndexError('Cannot choose from an empty sequence')
IndexError: Cannot choose from an empty sequence
[2018-06-23 13:55:54,511] [sbws.core.scanner] [ERROR] Unhandled exception caught while measuring foo: <class 'IndexError'> Cannot choose from an empty sequence
```
It seems we should check that there's at least 1 exit

juga0:
to perform usability tests. This will avoid raising IndexError
when choosing an exit.

Fixes #211

### DIFF
--- a/sbws/lib/destination.py
+++ b/sbws/lib/destination.py
@@ -186,6 +186,9 @@ class DestinationList:
             possible_exits = sorted(
                 possible_exits, key=lambda e: e.bandwidth, reverse=True)
             exits = possible_exits[0:num_keep]
+            if len(exits) < 1:
+                log.warning("There are no exits to perform usability tests.")
+                continue
             # Try three times to build a circuit to test this destination
             circ_id = None
             for _ in range(0, 3):


juga0:
> it may just mean none of them are capable of exiting to this destination's hostname:port
i'm using same destination (freebird) both in Tor network and test network, and this error is only happening in test network. So i don't think the problem is the destination not being available and instead the problem might be the exit policy or the way exits are choose.
I could check in the code, but you probably know: are using dirauths as exits when possible?. Asking because in testnet most (if not all) exits are dirauths.
I can  check later testnet and code and figure out, though maybe you have other idea.

pastly:
You're right that in your specific situation the issue is probably caused by a limited number of exits or an imperfect way of selecting exits.

My point is this:

Consider a user with 10 configured destinations. Consider what happens when sbws runs this code, determined that the first two destinations are usable, and is checking the third destination.

Suddenly for whatever reasons sbws couldn't find any exits to help test the third destination. With this code, sbws will stop testing destinations and return the first two. I don't think it should. If `continue` was used instead, it would mark the third as unusable and it would continue on to test the remaining 7 destinations.



----

> I could check in the code, but you probably know: are using dirauths as exits when possible?

It looks like using dirauths as exits when testing destinations is possible. We simply ask the RelayList for all relays with the Exit flag and don't remove dirauths.

juga0:
I figure out the problem in testnet~: there are only 2 relays with the Exit flag, and they are both in the same IP that the destination i'm using, all run by you :- )
I guess it is fine to solve this with ``continue``, and let it wait...
To be able to measure in the testnet, i'm going to add my dirauth the exit flag too, and use your other destination (that is not in the same IP, right?)

pastly:
My "tityos" destination is on a different IP address, correct.

juga0:
Right, changed, now working :)
Commit amended with ``continue``

PR #222, opened 2018-07-03
Issue192

### LINKED ISSUE #192: change assert is_controller_okay to log.warn and count the error
I think that we should not get that exception and instead log the error and count one more circuit error (if it is not already being count, i've not checked).
@pastly: if it easy for you, would you do it?, otherwise i'll check:

```
[2018-06-15 16:58:51,788] [sbws.core.scanner] [ERROR] Unhandled exception in worker thread
  File "/path/simple-bw-scanner/sbws/core/scanner.py", line 257, in dispatch_worker_thread
    return measure_relay(*a, **kw)
  File "/path/simple-bw-scanner/sbws/core/scanner.py", line 170, in measure_relay
    cb.controller, conf.getfloat('general', 'http_timeout'))
  File "/path/simple-bw-scanner/sbws/util/requests.py", line 7, in make_session
    socks_info = stem_utils.get_socks_info(controller)  
  File "/path/simple-bw-scanner/sbws/util/stem.py", line 197, in get_socks_info
    assert is_controller_okay(controller)
AssertionError
```

Also:

```
  File "/path/.virtualenvs/simplebwscanner3/bin/sbws", line 11, in <module>
    load_entry_point('sbws', 'console_scripts', 'sbws')()
  File "/path/simple-bw-scanner/sbws/sbws.py", line 51, in main
    exit(comm['f'](*comm['a'], **comm['kw']))
  File "/path/simple-bw-scanner/sbws/core/scanner.py", line 402, in main
    run_speedtest(args, conf)
  File "/path/simple-bw-scanner/sbws/core/scanner.py", line 362, in run_speedtest
    for target in rp.best_priority():
  File "/path/simple-bw-scanner/sbws/lib/relayprioritizer.py", line 49, in best_priority
    relays = copy.deepcopy(self.relay_list.relays)
  File "/path/simple-bw-scanner/sbws/lib/relaylist.py", line 132, in relays
    self._refresh()
  File "/path/simple-bw-scanner/sbws/lib/relaylist.py", line 175, in _refresh
    self._relays = self._init_relays()
  File "/path/simple-bw-scanner/sbws/lib/relaylist.py", line 169, in _init_relays
    assert stem_utils.is_controller_okay(c)
```

juga0:
Fixes #192.
It's catching all possible controller exceptions and logging them. 
If/when we know more details about why they happen, we might want to catch only those and do something different.
In some cases we might want to return different to what it's now returned.
This solution is almost the same as just letting the Exception raise.
It seems that we could also use ``add_status_listener`` [0] to the controller, though i'm not sure now which functions we should call in the different situations.

[0] https://stem.torproject.org/api/control.html#stem.control.BaseController.add_status_listener

### DIFF
--- a/sbws/lib/relaylist.py
+++ b/sbws/lib/relaylist.py
@@ -39,7 +39,7 @@ class Relay:
         else:
             try:
                 self._desc = cont.get_server_descriptor(fp, default=None)
-            except Exception as e:
+            except (DescriptorUnavailable, ControllerError) as e:
                 log.exception("Exception trying to get desc %s", e)
 
     def _from_desc(self, attr):
--- a/sbws/lib/relaylist.py
+++ b/sbws/lib/relaylist.py
@@ -32,6 +32,7 @@ class Relay:
                 self._ns = cont.get_network_status(fp, default=None)
             except Exception as e:
                 log.exception("Exception trying to get ns %s", e)
+                self._ns = None
         if desc is not None:
             assert isinstance(desc, ServerDescriptor)
             self._desc = desc
--- a/sbws/lib/relaylist.py
+++ b/sbws/lib/relaylist.py
@@ -40,7 +40,7 @@ class Relay:
             try:
                 self._desc = cont.get_server_descriptor(fp, default=None)
             except Exception as e:
-                log.exception("Exception trying to get ns %s", e)
+                log.exception("Exception trying to get desc %s", e)
 
     def _from_desc(self, attr):
         if not self._desc:
--- a/sbws/lib/relaylist.py
+++ b/sbws/lib/relaylist.py
@@ -1,6 +1,5 @@
 from stem.descriptor.router_status_entry import RouterStatusEntryV3
 from stem.descriptor.server_descriptor import ServerDescriptor
-import sbws.util.stem as stem_utils
 from stem import Flag
 from stem.util.connection import is_valid_ipv4_address
 from stem.util.connection import is_valid_ipv6_address
@@ -25,17 +24,22 @@ class Relay:
         '''
         assert isinstance(fp, str)
         assert len(fp) == 40
-        assert stem_utils.is_controller_okay(cont)
         if ns is not None:
             assert isinstance(ns, RouterStatusEntryV3)
             self._ns = ns
         else:
-            self._ns = cont.get_network_status(fp, default=None)
+            try:
+                self._ns = cont.get_network_status(fp, default=None)
+            except Exception as e:
+                log.exception("Exception trying to get ns %s", e)
         if desc is not None:
             assert isinstance(desc, ServerDescriptor)
             self._desc = desc
         else:
-            self._desc = cont.get_server_descriptor(fp, default=None)
+            try:
+                self._desc = cont.get_server_descriptor(fp, default=None)
+            except Exception as e:
+                log.exception("Exception trying to get ns %s", e)
 
     def _from_desc(self, attr):
         if not self._desc:
@@ -190,9 +194,12 @@ class RelayList:
 
     def _init_relays(self):
         c = self._controller
-        assert stem_utils.is_controller_okay(c)
-        relays = [Relay(ns.fingerprint, c, ns=ns)
-                  for ns in c.get_network_statuses()]
+        try:
+            relays = [Relay(ns.fingerprint, c, ns=ns)
+                      for ns in c.get_network_statuses()]
+        except Exception as e:
+            log.exception("Exception trying to init relays %s", e)
+            return []
         return relays
 
     def _refresh(self):
--- a/sbws/lib/relaylist.py
+++ b/sbws/lib/relaylist.py
@@ -1,6 +1,6 @@
 from stem.descriptor.router_status_entry import RouterStatusEntryV3
 from stem.descriptor.server_descriptor import ServerDescriptor
-from stem import Flag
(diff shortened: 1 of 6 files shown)


pastly:
First easy thing I noticed is that the changelog entry is in the wrong section. It should be in Unreleased, not 0.5.0. You'll have to add the `### Changed` section to Unreleased.

I'll still do a better review.

sbws/lib/relaylist.py
```diff
@@ -25,17 +24,22 @@ def __init__(self, fp, cont, ns=None, desc=None):
...
-        assert stem_utils.is_controller_okay(cont)
         if ns is not None:
             assert isinstance(ns, RouterStatusEntryV3)
             self._ns = ns
         else:
-            self._ns = cont.get_network_status(fp, default=None)
+            try:
+                self._ns = cont.get_network_status(fp, default=None)
+            except Exception as e:
+                log.exception("Exception trying to get ns %s", e)
```
pastly: If an exception is thrown, I'm not sure if `self` will end up with a `_ns` set to `None`, or if it will lack a `_ns` member. Please explicitly set `self._ns = None` if there was an exception.

sbws/lib/relaylist.py
```diff
@@ -25,17 +24,22 @@ def __init__(self, fp, cont, ns=None, desc=None):
...
         if desc is not None:
             assert isinstance(desc, ServerDescriptor)
             self._desc = desc
         else:
-            self._desc = cont.get_server_descriptor(fp, default=None)
+            try:
+                self._desc = cont.get_server_descriptor(fp, default=None)
+            except Exception as e:
+                log.exception("Exception trying to get ns %s", e)
```
pastly: Exception getting desc, not ns

sbws/util/stem.py
```diff
@@ -28,21 +27,25 @@ def closure_stream_event_listener(st):
             except (UnsatisfiableRequest, InvalidRequest) as e:
                 log.warning('Couldn\'t attach stream to circ %s: %s',
                             circ_id, e)
+            except Exception as e:
+                log.exception("Exception trying to get ns %s", e)
```
pastly: I'm pretty sure this message should be something else. Maybe 'Unknown exception while trying to attach stream {st.id} to circ {circ_id}: {e}' (but with real Python)

sbws/util/stem.py
```diff
@@ -72,9 +75,11 @@ def init_controller(port=None, path=None, set_custom_stream_settings=True):
 
 
 def is_bootstrapped(c):
-    if not is_controller_okay(c):
+    try:
+        line = c.get_info('status/bootstrap-phase')
+    except Exception as e:
+        log.exception("Exception bootstrapping %s", e)
```
pastly: We aren't bootstrapping in this function ... we are checking to see if we are bootstrapped. I'd say "Exception trying to check bootstrap phase" or something to that effect.

sbws/util/stem.py
```diff
@@ -178,25 +181,33 @@ def launch_tor(conf):
...
     cont.set_conf('__LeaveStreamsUnattached', '1')
-    log.info('Started and connected to Tor %s via %s', cont.get_version(),
-             conf['tor']['control_socket'])
-    return cont
+    try:
+        log.info('Started and connected to Tor %s via %s', cont.get_version(),
+                 conf['tor']['control_socket'])
+        return cont
+    except Exception as e:
+        log.exception("Exception trying to launch tor %s", e)
```
pastly: - If we get here, we didn't have an exception while trying to launch Tor. We got it while trying to query the control port for its version. So the error message is misleading.
- I don't think we need this try/except. We don't use a try/except every single time we use the controller (see how we don't wrap the `set_conf` calls just above this with try/except).

So I'd remove this try/except.

juga0: You're right that is not the correct place to catch it, and it should be in ``set_conf``, but i still think we should keep it, because it's the exception in which i run running 2 scanners with the same dir, and that mistake could happen to someone else.
I probably should add to the log something like: "Maybe other instance is using the same the same tor directory?"

juga0:
i wasn't happy with my own solution on catching all the possible exceptions, so i'm being less lazy, checking possible exceptions in https://stem.torproject.org/api/control.html, and catching only those. 
If we run into some other exception, then probably it means we need to modify something else.
But catching concrete exceptions, makes thing i should probably replace ``log.exception`` to some other level, since we know which is the exception, it's not needed to log the full trace.

juga0:
In https://github.com/pastly/simple-bw-scanner/blob/master/sbws/lib/circuitbuilder.py#L91, why did you add exceptions as ``CircuitExtensionFailed``?, according to https://stem.torproject.org/api/control.html, that's not there, though the documentation might not be complete.
I probably should remove https://github.com/pastly/simple-bw-scanner/pull/222/files#diff-814bf53a84b89545a05dd2761603efc4R96, since you're catching there possible exceptions for that method

juga0:
I did as i commented, please let me know if you prefer i leave ``Exception`` or if you prefer i squash commits or change something else.

pastly:
> i wasn't happy with my own solution on catching all the possible exceptions, so i'm being less lazy, checking possible exceptions

I'm glad. This is way better IMO.

> In [circuitbuilder.py when building a new circuit], why did you add exceptions as CircuitExtensionFailed?

Because stem's new_circuit calls extend_circuit which can raise that exception. I saw it in the wild and added it to the list of exceptions.

Issue #170, opened 2018-05-22
Unit tests not related with v3bw file header break when changing v3bw file header


juga0:
proposed by #169.
Possibly other time measures.
@pastly since you coded this part and know it better, would be fine for you to fix it once #169 is merged?.

teor2345:
Can you please explain this bug?

pastly:
From what I'm gathering, the changes @teor2345 and @juga0 have been discussing are only about how times are printed in the v3bw file and what they are labeled. That has no affect on how freshness is calculated when determining relay measurement priority.

juga0:
I changed the title of this issue since tests broken may not have to do anything with the timestamps nor the freshness.
However, i think that if when changing some part of the code affect to unit tests not related to that part, then either we need to refactor code or those unit tests are not unit tests.

pastly:
I see now. Okay. Yes the unit tests need some love and attention.

juga0:
The tests currently broken by issue169 branch are:
* https://github.com/pastly/simple-bw-scanner/compare/master...juga0:issue169?expand=1#diff-f8cd2c16d72d6b199edbe42c26e53671R76
* https://github.com/pastly/simple-bw-scanner/compare/master...juga0:issue169?expand=1#diff-dcea64797c457b7edf945e4bf3e04292R113
* https://github.com/pastly/simple-bw-scanner/compare/master...juga0:issue169?expand=1#diff-dcea64797c457b7edf945e4bf3e04292R147

juga0:
They can be fixed after issue169 branch is merged

pastly:
I wouldn't spend any more time trying to fix `tests/core/test_*.py`. They need overhauling, replacing, or removing.

juga0:
Hmm, this issue is now marked as 1.0 (must), so you think we should fix the tests for a MVP?.
I think i should fix #197 first, since many tests would break again after that change. Should i also add #197 to 1.0 (must)?

teor2345:
It seems like this issue was fixed in #225.

PR #225, opened 2018-07-05
Issue216: Revamp test support code

### LINKED ISSUE #209: Maybe unify integration and unit conftests

### LINKED ISSUE #216: unit tests: Refactor Results fixtures to reuse in resultdump and v3bwfile
As part of the big test refactoring plan, i'd start with the fixtures, in particular Results, since we use them in different tests.

I'd create:
1. fixture with result as dict, as in https://github.com/pastly/simple-bw-scanner/blob/master/tests/unit/lib/test_v3bwfile.py#L45
2. fixture with result as a class, reusing the previous one, as in https://github.com/pastly/simple-bw-scanner/blob/master/tests/unit/lib/test_results.py#L77, but without writing to disk
3. if fixture in disk is needed, i'd reuse the ones in the the commit https://github.com/pastly/simple-bw-scanner/pull/213/commits/ce0af44dace902a908534ce82015887d161a9ea6 and the previous one.

The fixtures in the previous commit, can be moved to a conftest.py one level up and be also reused by integration/confftest.py. Then (persistent)empty_dotsbws in unit and integration conftest.py can then be refactored too.

I already started to do it, but it's quite lot of work, so better ensure that @pastly  agrees with this.

I think this has to be in 1.0, but actually question, do we want 100% test coverage in a MVP?. It's quite lot of work, though i think it should be.

juga0:
Fixes #216
3 first commits would fix #209. The improvement is not that much.
It fixes part of #170, we could solve what is missing in other PRs.
Sorry for the big and not detailed commits.

### DIFF
--- a/sbws/core/stats.py
+++ b/sbws/core/stats.py
@@ -130,7 +130,7 @@ def print_stats(args, data):
         fastest_transfer/1024))
     print('Results come from', first, 'to', last, 'over a period of',
           duration)
-    if args.error_types:
+    if getattr(args, 'error_types', False) is True:
         _print_stats_error_types(data)
 
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,37 +1,72 @@
+"""pytest configuration for integration tests."""
+import argparse
 import pytest
-from tempfile import TemporaryDirectory
-from sbws.util.parser import create_parser
-from sbws.util.config import get_config
-from sbws.util.stem import launch_tor
-import sbws.core.init
 import os
 
+from sbws.core import init
+from sbws.lib.circuitbuilder import GapsCircuitBuilder as CB
+from sbws.lib.destination import DestinationList
+from sbws.lib.relaylist import RelayList
+from sbws.util.config import _get_default_config
+from sbws.util.stem import launch_tor
+
+
+class _PseudoArguments(argparse.Namespace):
+
+    """Just enough of the argparse.Namespace (what you get when you do
+    args = parser.parse_args()) to make get_config() happy
+
+    >>> args = _PseudoArguments(directory='/home/matt/.sbws')
+    >>> args.directory
+    '/home/matt/.sbws'
+
+    """
+
+    def __init__(self, **kw):
+        for key in kw:
+            setattr(self, key, kw[key])
+
 
 @pytest.fixture(scope='session')
-def parser():
-    return create_parser()
+def tmpdir(tmpdir_factory, request):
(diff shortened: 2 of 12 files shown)


juga0:
Ops, i still need to fix some tests

juga0:
now is ready for comments/merge

sbws/core/stats.py
```diff
@@ -130,7 +130,7 @@ def print_stats(args, data):
         fastest_transfer/1024))
     print('Results come from', first, 'to', last, 'over a period of',
           duration)
-    if args.error_types:
+    if getattr(args, 'error_types', False) is True:
```
pastly: I think we can avoid this change. See elsewhere for suggested change to _PseudoArguments.

tests/integration/conftest.py
```diff
@@ -1,44 +1,111 @@
...
 
+from sbws.core import init
+from sbws.lib.circuitbuilder import GapsCircuitBuilder as CB
+from sbws.lib.destination import DestinationList
+from sbws.lib.relaylist import RelayList
+from sbws.util.config import _get_default_config
+from sbws.util.stem import launch_tor
+
+
+class _PseudoArguments(argparse.Namespace):
```
pastly: I think we should add a function to this class that adds arbitrary attributes after the object has been created.

For example, when we want to test `sbws.core.stats` and call its `main()`, we can do the following in order to be able to avoid the `getattr()` and go back to `args.error_types`.

```
args.add_args(error_types=False)
```

sbws/core/stats.py
juga0: i changed that to ``if hasattr(args, 'error_types')`` before i pushed, but forgot to commit :)

tests/integration/conftest.py
juga0: hmm, i could add:
```
    def add_arg(self, arg):
        setattr(self, arg)
```
but then why just not using ``setattr`` 

In the tests i could just initialize the args with:
```
    args = _PseudoArguments(directory=sbwshome_dir, output=sbwshome_dir,
                            scale=False, log_level='debug', error_types=False)
```
no need for ``add_arg``
In non-tests, you ensure all the possible args are present with ``config.py``

pastly:
After thinking about it some more, I've decided I'm fine with the code the way it is. Merging.

PR #226, opened 2018-07-05
Add UML diagrams to doc

juga0:
Easy and more graphical


pastly:
Very nice. How can we regenerate these? Can it be done automatically somehow? It would be freaking sweet to do it when sphinx builds the site. It would be acceptable to have something in scripts/ to do it.

pastly:
> It would be freaking sweet to do it when sphinx builds the site.

We seem to be able to put arbitrary python in docs/source/conf.py and get sphinx to execute it when it runs. I tested this by adding

```
with open('foo.txt', 'wt') as fd:
    fd.write('hi\n')
```

And docs/source/foo.txt was created.

juga0:
Yes, it can be done automatically. The only problem i've experienced before is that it can not be included in a distribution package (specially Debian) because the XML generated is not deterministic.
Let me add the script...

juga0:
The last commit should work with  ``cd docs; make umlsvg``, but is not now. Will check tomorrow

juga0:
Bad bash var replacement. Now should work

setup.py
```diff
@@ -82,6 +82,6 @@ def find_version():
         'dev': ['flake8'],
         'test': ['tox', 'pytest', 'coverage'],
         # recommonmark: to make sphinx render markdown
-        'doc': ['sphinx', 'recommonmark'],
+        'doc': ['sphinx', 'recommonmark', 'pylint'],
```
pastly: Based on using pyreverse in the Makefile, I'm guessing this is supposed to be pyreverse, not pylint.

Could easily be wrong.

Issue #214, opened 2018-06-25
Add optional logging to syslog


pastly:
We currently maybe support logging to syslog. There's lines in config.log.default.ini that suggest to me that it might work, but I've never tested it. And if I don't test things for a long period of time, they tend to stop working :)

After #186 is fixed, we will no longer even pretend to support logging to syslog. So if we want to support that, we will need to add it back.

pastly:
Closed via c0e6945045ab6f77d29982278e761695706ae200..0306526e2a83968d3383bf8df5c06b41c0ee71ab

PR #230, opened 2018-07-07
Ticket26683

juga0:


juga0:
Didn't realize we already have #214.
Fixes #214.

sbws/config.default.ini
```diff
@@ -87,6 +87,8 @@ min_relays = 50
 to_file = yes
 # Whether or not to log to stdout
 to_stdout = no
+# Whether or not to log to syslog
+to_syslog = no
```
pastly: Please add to_syslog_level defaulting to `${level}`.

Please add to_syslog_format (feel free to have it default to the string you like instead of `${format}`

sbws/config.log.default.ini
```diff
@@ -27,20 +27,18 @@ class = handlers.RotatingFileHandler
...
-#[handler_to_syslog]
-#class=handlers.SysLogHandler
-#formatter=to_syslog
-#level = INFO
-#args = ()
+# for logging to system log
+[handler_to_syslog]
+class=handlers.SysLogHandler
+formatter=to_syslog
+level = INFO
```
pastly: Please don't set the level here so it is more straightforward to understand what messages will (not) be ignored.

sbws/config.log.default.ini
```diff
@@ -27,20 +27,18 @@ class = handlers.RotatingFileHandler
...
-## syslog-like formater
-#[formatter_to_syslog]
-## hostname should be added here with a context filter, dunno if there is a way
-## to add it here
-#format = %(asctime)s %(module)s[%(process)s]: <%(levelname)s> %(message)s
-#datefmt = %b %d %H:%M:%S
+# syslog-like formatter
+[formatter_to_syslog]
+format = %(asctime)s %(module)s[%(process)s]: <%(levelname)s> %(message)s
+datefmt = %b %d %H:%M:%S
```
pastly: See previous commit about letting the user set the format with `to_syslog_format`.

If it isn't necessary to set the date format in order to send messages to the syslog, can we not set it? I've never tried logging to syslog, but I would expect it to set the date format as it wants and not rely on us. But I really don't know.

This section can most likely be empty like the other formatters.

sbws/config.default.ini
juga0: i added then removed before opening this PR cause thought those settings doesn't probably need to be modified by users, however i'm not sure and it's fine for me to add them

sbws/config.log.default.ini
juga0: ok, as commented above

sbws/config.log.default.ini
juga0: > If it isn't necessary to set the date format in order to send messages to the syslog, can we not set it?

it wasn't the case time ago, but i'll check, i think you're right and that's not needed

juga0:
Yes i tested (and worked), will do the modifications you suggest

juga0:
Made the changes you suggest

PR #231, opened 2018-07-10
Ticket26740

juga0:


cron.d/sbws
```diff
@@ -0,0 +1,19 @@
...
+# 5-15 or 35-45 minutes past the hour
+# The following commented lines are all the possible times at which sbws 
+# should run. Choose only one.
+# 5 * * * *
+# 10 * * * *
+# 15 * * * *
+# 35 * * * *
+# 40 * * * *
+# 45 * * * *
+40 * * * *    $SBWSBIN -d $SBWSHOME generate
```
teor2345: Do crontab files need to end in a newline?

pastly: I'll merge as soon as @juga0 answers and, if necessary, fixes.

juga0: not AFAIK, but i did not find documentation about it that confirms it.
I tested it works.

juga0:
> I tested it works.

With the last commit, replace $HOME

pastly:
Don't system packages usually put their homes in /var/lib? 

If $HOME is set in the cron environment, wouldn't it be better to use $HOME so we don't have to remember to update that variable in this file?

juga0:
> Don't system packages usually put their homes in /var/lib?

yes, though since `sbws` does not require root to run, i was creating having in mind that ``make install`` could be run by a user without root privileges.
However, while there's i don't add a generic ``Makefile`` that accepts ``prefix``, and can be used standalone or with system packages, this ``crontatb`` files doesn't make much sense.
So probably we should close this as ``wontfix`` and the system package would take care of system configurations/files, or i should create a ``Makefile`` at least that manages the system files.
Will think about it a bit more.

juga0:
Closing because of what commented above.
Instead document generate in PR #235

PR #232, opened 2018-07-11
Log sbws version info line on startup

pastly:
Produces a line such as the following.

`[2018-07-11 11:27:35,464] [sbws.sbws] [INFO] sbws 0.6.1-dev with python 3.6.5 on Darwin-15.6.0-x86_64-i386-64bit, stem 1.6.0-dev, and requests 2.18.4`

### DIFF
--- a/sbws/sbws.py
+++ b/sbws/sbws.py
@@ -7,6 +7,10 @@ from sbws.util.config import get_config
 from sbws.util.config import validate_config
 from sbws.util.config import configure_logging
 from sbws.util.parser import create_parser
+from sbws import __version__ as SBWS_VERSION
+from stem import __version__ as STEM_VERSION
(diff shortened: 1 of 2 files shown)


sbws/sbws.py
```diff
@@ -7,6 +7,10 @@
 from sbws.util.config import validate_config
 from sbws.util.config import configure_logging
 from sbws.util.parser import create_parser
+from sbws import __version__ as SBWS_VERSION
+from stem import __version__ as STEM_VERSION
+from requests.__version__ import __version__ as REQ_VERSION
```
juga0: i'd just not use capitals here, since __version__ is not in capitals.
I'd use ``as version`` for ``sbws`` version and ``stem_version``, ``requests_version``

pastly:
Made variable name changes as requested and merged.

PR #234, opened 2018-07-13
Ticket26736

juga0:


tests/unit/lib/test_v3bwfile.py
```diff
@@ -145,4 +119,15 @@ def test_from_arg_results_write(datadir, tmpdir, conf, args):
     v3bwfile = V3BWFile.from_arg_results(args, conf, results)
     output = os.path.join(args.output, now_fname())
     v3bwfile.write(output)
+    print(output)
```
pastly: This looks like a debugging print statement. Do we need it?

If you push a fixup commit, I'll squash and merge!

juga0: No, we don't need it, it was indeed for debugging.
My editor is not reloading files when i change branch, that's why i'm not realizing about the lines i use for my -dev branch remain in the one i use to push

teor2345:
> My editor is not reloading files when i change branch, that's why i'm not realizing about the lines i use for my -dev branch remain in the one i use to push

You might find “git add -p” helpful.
It shows you the changes that you are about to add.
(But it doesn’t add new files.)

PR #236, opened 2018-07-16
Add state file with scanner start time

### LINKED ISSUE #166: Add state file, as well as functions for safely getting/setting data in it
JSON encoded.

@juga0, if I do this, I'd like to replace your started_at file with a field in the state file. Is that okay?

Initially, it would look something like

```
$ cat ~/.sbws/state
{
    started_at: 1526920769
}
```

pastly:
Closes #166.

### DIFF
--- a/sbws/lib/v3bwfile.py
+++ b/sbws/lib/v3bwfile.py
@@ -190,6 +190,10 @@ class V3BWHeader(object):
 
     @staticmethod
     def generator_started_from_file(conf):
+        '''
+        ISO formatted timestamp for the time when the scanner process most
+        recently started.
+        '''
         state = State(conf['paths']['state_fname'])
         if 'scanner_started' in state:
             return state['scanner_started']
--- a/sbws/lib/v3bwfile.py
+++ b/sbws/lib/v3bwfile.py
@@ -9,8 +9,9 @@ from statistics import median
 from sbws import __version__
 from sbws.globals import SPEC_VERSION, BW_LINE_SIZE
 from sbws.lib.resultdump import ResultSuccess, _ResultType
-from sbws.util.filelock import FileLock, DirectoryLock
+from sbws.util.filelock import DirectoryLock
 from sbws.util.timestamp import now_isodt_str, unixts_to_isodt_str
+from sbws.util.state import State
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


sbws/lib/v3bwfile.py
```diff
@@ -210,7 +190,11 @@ def num_lines(self):
 
     @staticmethod
     def generator_started_from_file(conf):
-        return read_started_ts(conf)
+        state = State(conf['paths']['state_fname'])
```
juga0: I'd just add here a comment (or docstring) about what is supposed to be read here.
Something like "ISO formated timestamp which represents the date and time...."

sbws/util/state.py
```diff
@@ -0,0 +1,62 @@
+from sbws.util.filelock import FileLock
+import os
+import json
+
+
+class State:
+    _ALLOWED_TYPES = (int, float, str, bool, type(None))
```
juga0: i'd add also here a docstring or comment about what is this class for and which format it's supposed to store/read (json)

juga0:
Regarding ``datadir`` it could be named ``raw_bw_measurements``/``results`` in the fs or something like that.
In the code it could be renamed ``results_dname``, but we can also leave it like this.

juga0:
i'm wondering how easy would be to inherit from json class and overwrite methods. read/write could add just the lock.
It seems that this could also be reaused for results?.
If it means lot of refactoring, it's probably not worth to change this now.

tests/unit/util/test_state.py
```diff
@@ -0,0 +1,120 @@
...
+    del d['a']
+    del state['a']
+    assert len(state) == len(d)
+    for key in d:
+        assert d[key] == state[key]
+
+    attempt_keys = (15983, None, True, -1.2, [], {}, set())
+    for key in attempt_keys:
+        try:
+            del state[False]
```
pastly: This should be `state[key]` not `state[False]`

tests/unit/util/test_state.py
```diff
@@ -0,0 +1,120 @@
...
+
+def test_state_iter(tmpdir):
+    state = State(os.path.join(str(tmpdir), 'statefoo'))
+    for key in state:
+        pass
+    d = {'a': 1, 'b': 2, 'c': 3, 'd': 4}
+    for key in d:
+        state[key] = d[key]
+    for key in state:
+        pass
```
pastly: This should actually test something. Consider checking the length of the iter, as well as making sure the keys in the iter are the same as the keys in `d`.

pastly:
After the above reviews, I plan on:

- leaving "datadir" alone for now as there is a ticket for it
- adding the two docstrings @juga0 suggested
- making the test fixes I suggested
- **not** trying to inherit JSON class(es) because
    - while results are stored as JSON, they're stored as one JSON object per line to facilitate easier writing and easier ignoring of malformed result data. This is not compatible with the State class in this branch, and it isn't immediately obvious how I would change the State class to work with this design choice.
    - I skimmed the JSON encoding/decoding source code in cpython and it wasn't immediately obvious what I should reimplement and build upon in order to achieve this syncing with the disk functionality.

pastly:
Done with the above mentioned changes. If you approve, let me know and I can resolve the conflict in README.md and merge. Or if you want to, you can.

juga0:
i tend to over-engineer things and i think this is a bit doing it, but it's already done, so it's fine
Re. README, yes, you can solve the conflict

pastly:
Merged in ccba34a43cf1860d2792deb5cff20508f5c6d848, thanks for the review @juga0

PR #238, opened 2018-07-18
Ticket26862: config file argument

juga0:


sbws/util/config.py
```diff
@@ -21,6 +25,12 @@
 log = logging.getLogger(__name__)
 
 
+def _create_user_config_file(fname):
+    """Copy minimal user config to user config path."""
+    if not os.path.isfile(fname):
+        shutil.copyfile(MINIMUM_USER_CONFIG_PATH, fname)
```
pastly: Do you think we should assert on `fname` not existing here? I think so. 

If the file exists already, we don't create/overwrite anything. We silently do nothing. As a programmer, I'd expect a function with "create" in its name to do *something* when the file already exists. I would expect it to do (at least) one of the following.

- overwrite it
- warn about it existing (and about overwriting it, if it does so)
- assert that the file doesn't exist

I would prefer the assert. If not that, then a different function name. If not that, a debug log warning that it exists.

sbws/util/config.py
```diff
@@ -29,52 +39,57 @@ def _read_config_file(conf, fname):
     return conf
 
 
+def _extend_config(conf, fpath):
+    """Extend ConfigParser from file configuration."""
+    conf = _read_config_file(conf, fpath)
+    return conf
+
```
pastly: Unless I'm missing something, `_extend_config()` is an unnecessary wrapper around `_read_config_file()`, thus every time `_extend_config()` is called, `_read_config_file()` could be called instead.

If you like the name `_extend_config()` better (I do), I'd just rename `_read_config_file()` to that.

sbws/util/parser.py
```diff
@@ -22,12 +21,11 @@ def create_parser():
     p.add_argument('--log-level',
                    choices=['debug', 'info', 'warning', 'error', 'critical'],
                    help='Override the sbws log level')
-    p.add_argument('-d', '--directory', default=_default_dot_sbws_dname(),
-                   help='Name of the .sbws directory')
+    p.add_argument('-c', '--config',
+                   help='Path the sbws config file')
```
pastly: "Path **to** the sbws config file"

sbws/util/config.py
```diff
@@ -29,52 +39,57 @@ def _read_config_file(conf, fname):
...
-def _get_default_logging_config(args, conf=None):
+    if args.config:
+        if not os.path.isfile(args.config):
+            log.warning('Configuration file %s not found, '
+                        'using default configuration, and creating a minimal '
+                        'configuration in %s.', args.config, args.config)
+            _create_user_config_file(args.config)
+            return conf
+        return _extend_config(conf, args.config)
+    _create_user_config_file(USER_CONFIG_PATH)
```
pastly: If we make `_create_user_config_file()` assert on the file not existing, then we should only conditionally call this function here.

tests/integration/conftest.py
```diff
@@ -80,7 +79,6 @@ def conf(sbwshome_dir):
 
 @pytest.fixture(scope='session')
 def sbwshome(sbwshome_dir, args, conf):
-    init.main(args, conf)
     return sbwshome_dir
```
pastly: This function looks unnecessary. Instead of using this fixture, why not just use the parent `sbwshome_dir` fixture in tests?

sbws/sbws.py
```diff
@@ -16,6 +17,13 @@
 log = logging.getLogger(__name__)
 
 
+def _ensure_dirs(conf):
+    log.debug('Ensuring all dirs exits.')
+    os.makedirs(conf['paths']['datadir'])
+    os.makedirs(conf['paths']['v3bw'])
+    os.makedirs(conf['paths']['log'])
+
```
pastly: This needs to have `exist_ok=True` for each.

pastly: `v3bw_dname` and `log_dname`

juga0:
i made the changes you requested though in a bit different way.
Let me know if more explanations or additional changes are needed.

juga0:
Thinking it better, if we allow the config paths to have env vars or ``~``, then all paths should be expanded (in case the user choose paths not relative to sbws_home).
I think logic would be simpler not allowing config paths nor env vars for now (we are not taking env vars into account anywhere else in the config). Then ``sbws_home`` can be set to ``~`` by default when not provided in the config file. Opinions?

pastly:
> Thinking it better, if we allow the config paths to have env vars or ~, then all paths should be expanded (in case the user choose paths not relative to sbws_home).

This is my second suggested fix. "Here's another possible fix: after getting the complete config in get_config(), call something that expands `sbws_home` (**or all the paths in the config**)." My second suggested fix was also my least favorite.

> I think logic would be simpler not allowing config paths to have ~ nor env vars for now

I would agree, but then you made this PR that adds a`~` in `sbws_home` and I convinced myself that allowing `~` would be important for flexible software that lends itself to packaging. 

I think the beautiful solution keeps logic very simple. In fact, it might be simpler since no paths are special cases and they all get handled the same way automatically.

Define the following code (but turned into real Python)

```
# Add in sbws/util/config.py
def _parse_path(path):
    return os.path.expanduser(os.path.expandvars(path))

# And everywhere we create a ConfigParser, do it this way
conf = ConfigParser(interpolation=ExtendedInterpolation(), converters={'path': _parse_path})
# (ExtendedInterpolation() is not new, just the coverters argument)

#############

# And everywhere we get a path from the config, do it this way.
# Just like we get ints with .getint() and bools with .getboolean()
some_fname = conf.getpath('foo', 'some_fname')
```

sbws/util/config.py
```diff
@@ -21,57 +25,87 @@
...
-    return conf
-
-
-def get_user_example_config():
-    conf = ConfigParser(interpolation=ExtendedInterpolation())
-    fname = os.path.join(PKG_DIR, 'config.example.ini')
-    assert os.path.isfile(fname)
-    conf = _read_config_file(conf, fname)
+    # it is only needed to expand user and vars to have the rest of the
+    # paths correctly expanded too
```
pastly: To elaborate a little more on why I like the "beautiful solution" the most: this comment is true for the default configuration options as they exist today and is true for how I expect users to use sbws. But other than that, it isn't necessarily true. I don't think there's anything stopping users from configuring paths in such a way that this comment becomes false.

juga0:
i used the "beautiful fix" you suggested.

After thinking more about, i changed again the function to create the user config. I think we should not create any user config by default, as when using system packages it would be overwritten by the system.
Instead, i moved back the example config to examples, so that a user can create a user config from it easily.
I also changed some default tor directories, as at some point, in a system package, those directories would be managed by a service and would be in different locations.
And i also changed the logging to stdout by default, as that's would a system service would expect.

juga0:
meh, it seems that in python 3.4 ConfigParser does not accept converters...

pastly:
@juga0, how do you feel about dropping support for Python 3.4? [Stretch has 3.5](https://packages.debian.org/stretch/python3) as [does Xenial](https://packages.ubuntu.com/xenial/python3). People on other distros (1) will not have a nice .deb package to install and will have to git clone or pip install, (2) can use [pyenv](https://github.com/pyenv/pyenv) to get a newer python.

sbws/util/config.py
```diff
@@ -21,60 +24,67 @@
...
+                        converters={'path': _expand_path})
+    return _extend_config(conf, DEFAULT_CONFIG_PATH)
 
 
 def _get_user_config(args, conf=None):
+    """Get user configuration.
+    If a custom path is passed, search for user configuration there.
+    In any case, create a minimal user configuration in user configuration path
+    , in case it needs to be overwritten.
+    """
```
pastly: This function doesn't currently create a file. I think that just means the comment needs to be updated.

docs/source/sbws.core.rst
```diff
@@ -1,54 +0,0 @@
...
-    :show-inheritance:
-
-
-Module contents
----------------
-
-.. automodule:: sbws.core
-    :members:
-    :undoc-members:
-    :show-inheritance:
```
pastly: Did you mean to delete this file?

docs/source/config.rst
```diff
@@ -1,12 +1,25 @@
...
+How to create user customized configuration files
+--------------------------------------------------
+``sbws`` use internal configuration files.
+These are **not** intented to be modified by a user.
+If a user needs a customized configuration file, the user needs to create it.
+There's an example configuration file with the minimum data that a user
+probably needs to edit in ``examples``. For more advanced configuration options see
+documentation below.
+``sbws`` will check by default whether there's a user customized file in
+``~/sbws.ini`` and use it when it exists.
```
pastly: ~/.sbws.ini (missing the dot)

juga0:
> @juga0, how do you feel about dropping support for Python 3.4?

up for it. There're has been already a couple of tickets where i was also asking this since giving support for Python 3.4 required extra work

juga0:
Thanks for pointing out the mistakes. They should be fixed now

juga0:
When you're happy with the changes, let me know and i can rebase to master and fix the conflicting files problems

juga0:
Continuing in #242 cause of rebase

juga0:
For the next time, let's try to merge old PRs before creating new features directly into master.

pastly:
Merged via #242

PR #240, opened 2018-07-25
Ticket26926: man pages

juga0:


pastly:
Is there any way we can auto-generate one (or both) of these? I don't want to have to remember to update them when we add a new config option or argument.

In anticipation of the answer being "no, not really", I'm merging this. Thanks @juga0.

juga0:
> Is there any way we can auto-generate one (or both) of  these? 

which "both" do you mean?

> I don't want to have to remember to update them when we add a new  config option or argument. In anticipation of the answer being "no, not really", 

Do you mean to update the man for the config file?. Yeah, "not really". 
The man for the binary could be created with the --help, but then it would not have the other typical sections.

pastly:
> Is there any way we can auto-generate one (or both) of these?

"both" means the both man_sbws.rst and man_sbws_ini.rst (or whatever their exact file names are)

The man for the config file is what I'm mainly thinking about. There's a lot of options. I imagine I could write a parsing script that could take the following content from config.default.ini and turn it into man page content, but that's potentially a lot of work and we have better things to be doing.

```
[general]
# TYPE: int
# DESCRIPTION: Days into the past that measurements are considered valid.
data_period = 5
```

Would become

```
general
  data_period = INT
    Days into the past that measurements are considered valid. (Default: 5)
```

PR #241, opened 2018-07-25
Ticket26937: check for free disk space

juga0:


sbws/util/fs.py
```diff
@@ -0,0 +1,50 @@
...
+def sbws_required_disk_space(conf):
+    """Disk space required by sbws files.
+    Rough calculations.
+
+    :param ConfigParser conf: sbws configuration
+    :returns: int, size in MiB
+    """
+    # Number of relays per line average size in Bytes
+    size_v3bw_file = 7500 * 220
+    num_v3bw_files = int(conf['general']['data_period'])
```
pastly: data_period is the number of days into the past that we consider measurements valid. Why does this line assume that the number of v3bw files will be the same as this value?

sbws/util/fs.py
```diff
@@ -0,0 +1,50 @@
...
+    # not counting compressed files
+    space_v3bw_files = size_v3bw_file * num_v3bw_files
+    # not counted rotated files and assuming that when it is not rotated the
+    # size will be aproximately 10MiB
+    size_log_file = (int(conf['logging']['to_file_max_bytes']) or 10485760) \
+        if conf['logging']['to_stdout'] == 'yes' else 0
```
pastly: Shouldn't this be to_file, not to_stdout? Also, you should do `if conf.getboolean('logging', 'to_file')` since there are more was to indicate a true value than the word "yes" https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.getboolean

sbws/util/fs.py
```diff
@@ -0,0 +1,50 @@
...
+    :returns: int, size in MiB
+    """
+    # Number of relays per line average size in Bytes
+    size_v3bw_file = 7500 * 220
+    num_v3bw_files = int(conf['general']['data_period'])
+    # not counting compressed files
+    space_v3bw_files = size_v3bw_file * num_v3bw_files
+    # not counted rotated files and assuming that when it is not rotated the
+    # size will be aproximately 10MiB
+    size_log_file = (int(conf['logging']['to_file_max_bytes']) or 10485760) \
```
pastly: `conf.getint('logging', 'to_file_max_bytes')` instead of wrapping with `int()`

Since 10 MiB is specified in the config.default.ini and that file should not be edited by anyone, I do not think we should hardcode a fallback of 10 MiB here

sbws/util/fs.py
```diff
@@ -0,0 +1,50 @@
...
+    # roughly...
+    space_result_files = space_v3bw_files
+    # duplicate everything to warn early
+    size_total = (space_v3bw_files + size_log_file + space_result_files) * 2
+    size_total_mb = round(size_total / (1024 ** 2))
+    return size_total_mb
+
+
+def df(path):
+    """Return space left on device where path is."""
```
pastly: Please note that the units are MiB

sbws/util/fs.py
```diff
@@ -0,0 +1,50 @@
...
+    size_v3bw_file = 7500 * 220
+    num_v3bw_files = int(conf['general']['data_period'])
+    # not counting compressed files
+    space_v3bw_files = size_v3bw_file * num_v3bw_files
+    # not counted rotated files and assuming that when it is not rotated the
+    # size will be aproximately 10MiB
+    size_log_file = (int(conf['logging']['to_file_max_bytes']) or 10485760) \
+        if conf['logging']['to_stdout'] == 'yes' else 0
+    # roughly...
+    space_result_files = space_v3bw_files
```
pastly: How do we know this will be close to true?

sbws/util/fs.py
```diff
@@ -0,0 +1,50 @@
...
+
+def is_low_space(conf):
+    """Warn and return True when the space left on the device is less than
+    what is needed for sbws and False otherwise needs.
+    """
+    disk_required_mb = sbws_required_disk_space(conf)
+    disk_avail_mb = df(conf['paths']['sbws_home'])
+    if disk_avail_mb < disk_required_mb:
+        log.warn("The space left on the device (%s MiB) is less than "
+                 "the minimum recommented to run sbws (%s MiB)."
```
pastly: "recommended"

sbws/util/fs.py
juga0: I think that i intended to calculate size of the minimum number of results files to keep, not v3bw files.
For v3bw files, it would be enough to keep only the last 2 ones that should have been generated in the last 2 hours. 
This is going to depend a lot on the crontab that will generate/clean files and would be provided separately, but at least this would give an estimation on the minimum space required

sbws/util/fs.py
juga0: you're right, mistake

sbws/util/fs.py
juga0: good point

sbws/util/fs.py
juga0: not sure now why i did that :), probably i should calculate the space they take separately, in the way commented in https://github.com/pastly/simple-bw-scanner/pull/241#discussion_r207324143

sbws/util/fs.py
juga0: makes sense

juga0:
> I also don't know why two specific log messages were picked as places to catch an exception when we think that maybe we ran out of disk space. 

In concrete cases where i run out of space, they happened there.

> Couldn't this happen every time we log? 

Yes

> Shouldn't we therefore catch this exception at every log message? (Please do not write code to do that)

Maybe, but we would need to change all logs, so probably we do not want that.

What about just leaving the logs in this PR?

juga0:
Fixed the issues you pointed out when calculating the space

sbws/util/fs.py
```diff
@@ -0,0 +1,58 @@
...
+    # ~4550 is the number of lines of the biggest result file
+    size_result_file = 4550 * 1000
+    num_result_files = int(conf['general']['data_period'])
+    # not counting compressed files
+    space_v3bw_files = size_v3bw_file * num_v3bw_files
+    space_result_files = size_result_file * num_result_files
+    # not counted rotated files and assuming that when it is not rotated the
+    # size will be aproximately 10MiB
+    size_log_file = conf.getint('logging', 'to_file_max_bytes') or 10485760 \
+        if conf.getboolean('logging', 'to_stdout') else 0
```
pastly: This is supposed to be `to_file`.

PR #244, opened 2018-08-28
Ticket27342: poor error handling (env vars in config is not okay)

pastly:
I'm embarrassed we even let this happen. Sbws would not start for me. See commit messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Stop pretending we allow environment variables in the config file, and stop
+  trying to use one. (#27342)
+
 ## [0.7.0] - 2018-08-09
 
 **Important changes**:
--- a/sbws/config.default.ini
+++ b/sbws/config.default.ini
@@ -59,7 +59,7 @@ max_download_size = 1073741824
 [tor]
 datadir = ${paths:sbws_home}/tor
 # create runtime dir for data that does not need to be persistent
-run_dpath = $XDG_RUNTIME_DIR/sbws
+run_dpath = ${tor:datadir}/run
 control_socket = ${tor:run_dpath}/control
 pid = ${tor:run_dpath}/tor.pid
 # note this is a directory
--- a/sbws/core/scanner.py
+++ b/sbws/core/scanner.py
@@ -379,7 +379,7 @@ def main(args, conf):
 
     os.makedirs(conf.getpath('paths', 'datadir'), exist_ok=True)
 
-    state = State(conf['paths']['state_fname'])
+    state = State(conf.getpath('paths', 'state_fname'))
     state['scanner_started'] = now_isodt_str()
 
     try:
--- a/sbws/util/config.py
+++ b/sbws/util/config.py
@@ -25,10 +25,10 @@ log = logging.getLogger(__name__)
 
 
 def _expand_path(path):
-    """Expand path string containing shell variables and ~ constructions
-    into their values.
-    """
-    return os.path.expanduser(os.path.expandvars(path))
+    """ Wrapper for os.path.expanduser. If we ever want to support expanding
+    environment variables in config values, this is where they should be
+    expanded. """
+    return os.path.expanduser(path)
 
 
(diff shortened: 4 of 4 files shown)


teor2345:
How did the tox tests pass with these errors in the config?
Can you update the travis CI to avoid issues like this in future?

pastly:
Yes I think that should be possible and straight forward. I'll add a test that makes sure an environment variable isn't interpolated in the config (less interesting) and a test that tries launching sbws from scratch in a new processes and making sure it runs as expected.

pastly:
>  i'm still getting the error with this branch.

How can I reproduce the error you are seeing?

> The way you change config.default.ini is fine, but would you mind to add something like [... code ...] in the moment when that is added to the config, so we can still use env var when it's set?.

I don't want to support this environment variable as a special case.

juga0:
> How can I reproduce the error you are seeing?

```
rm -rf ~/.sbws
sbws scanner
[error]
ls ~/.sbws
datadir  log  v3bw
git branch -l
* ticket27342
```

> I don't want to support this environment variable as a special case.

That's the default case in system packages (Debian).
A service that runs as a non-root user put run time files in ``/run/theuser/``
A service that runs as a root user put the run time files in ``/var/run``.
``tor`` service put its run time files in ``/var/run/tor``.
Our ``tor`` directory should also put their files run time directories.

juga0:
> How can I reproduce the error you are seeing?

oh, sorry, that only happens in a chroot, so it's fine.

juga0:
I'd improve the message:
```
[2018-08-28 13:40:45,836] [sbws.globals] [CRITICAL] No enabled destinations in config
```

To say something like ``You nee to configure destinations, please create a <thehomepaththatsbwsisusing>/config.ini or specify the location of the config.ini with the the option -c. See man sbws.ini to know how to configure destinations``

pastly:
> reproduce error.

I get the following, which is correct and expected.

```
(venv-test) matt@spacecow:~/src/simple-bw-scanner$ sbws scanner
[2018-08-28 09:40:59,911] [sbws.sbws] [INFO] sbws 0.7.1-dev with python 3.5.3 on Linux-4.9.0-8-amd64-x86_64-with-debian-9.5, stem 1.6.0-dev, and requests 2.19.1
[2018-08-28 09:41:31,311] [sbws.util.stem] [INFO] Started and connected to Tor 0.3.3.9 (git-ca1a436fa8e53a32) via /home/matt/.sbws/tor/run/control
[2018-08-28 09:41:36,199] [sbws.globals] [CRITICAL] No enabled destinations in config
[2018-08-28 09:41:36,256] [sbws.lib.resultdump] [WARNING] Results files that are valid not found. Probably sbws scanner was not run first or it ran more than 7 days ago or it was using a different datadir than /home/matt/.sbws/datadir.
```

> Supporting env vars, but correct place in Debian is to put them in [places]

When creating the Debian package, you can include a config file that sets these paths to be the exact right places, right?

juga0:
> When creating the Debian package, you can include a config file that sets these paths to be the exact right places, right?

i do not include any config file in the debian package and that is intentional, since updates should not modify files that the user might modify.

The commit message in 79fdb92aafa728db1bcd1a9294ddb4afde3efd0c is not correct.

ExtendedInterpolation does support environment variables. 

And the traceback says the error:

```
configparser.InterpolationSyntaxError: '$' must be followed by '$' or '{', found: '$HOME/.sbws'
```

The only change required is s/$/$$/:

``run_dpath = $$XDG_RUNTIME_DIR/sbws``

Changing that i get:

```
[sbws.util.stem] [INFO] Started and connected to Tor 0.2.9.15 (git-2dc1a1a2abab5403) via /run/user/1000/sbws/control
```

pastly:
That's great news and means I can undo just about everything I did, thanks.

juga0:
i would keep the original title for this message ``poor error handling``

``env vars in config is not okay`` is not correct

juga0:
Commit 6c00e2a1194b163ee6eccf057372bc3c125aa227 should remain.

We should still change the info message to say what the user should do

pastly:
> We should still change the info message to say what the user should do

See https://trac.torproject.org/projects/tor/ticket/27358 for that.

Closing this in favor of that ticket, plus one I will create for making a test that launches sbws from the command line to hopefully catch something like this in the future.

PR #245, opened 2018-08-28
Ticket27341

juga0:
sbws install doc is confusing


juga0:
> f you are doing this because I wrote "Strive to write many small commits each containing an atomic change instead of one large mega-commit." in CONTRIBUTING

Yes, and i started to liked :)

See https://github.com/pastly/simple-bw-scanner/pull/215#issuecomment-401590213. I've not contributed to CONTRIBUTING to focus on having a MVP.

I'm not usually happy with my commit messages, but it would be hard to accept any contribution from me if "nice" commit messages are required from me.

PR #261, opened 2018-10-05
Add types to new generate arguments

pastly:
--secs-recent is a float to support weird people who might want to
include fractional seconds
--secs-away is a float for the same reason
--min-num is an int because there's no such thing as half a result


juga0:
Fixed by https://github.com/torproject/sbws/pull/267/ using int

PR #262, opened 2018-10-05
Remove default values from help strings

pastly:
Because

1. They might become incorrect if the constants change in the future
2. (The main reason) They are already printed because the argparser is
an ArgumentDefaultsHelpFormatter type


juga0:
Fixed by #265

PR #273, opened 2018-10-15
Ticket28045 support python37

juga0:


juga0:
Travis still tries to download python from trusty.
Closing this for now

PR #278, opened 2018-10-22
Don't assume we can resolve a domain name

pastly:
Even for a valid domain that usually resolves, sometimes the local DNS
resolver won't be able to resolve it (duh). Catch the exception that gets
thrown in that instance and fail gracefully by returning an empty list.

### DIFF
--- a/sbws/globals.py
+++ b/sbws/globals.py
@@ -67,8 +67,16 @@ def touch_file(fname, times=None):
 
 def resolve(hostname, ipv4_only=False, ipv6_only=False):
     assert not (ipv4_only and ipv6_only)
+    results = []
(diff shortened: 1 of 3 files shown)


sbws/globals.py
```diff
@@ -67,8 +67,16 @@ def touch_file(fname, times=None):
 
 def resolve(hostname, ipv4_only=False, ipv6_only=False):
     assert not (ipv4_only and ipv6_only)
+    results = []
+    try:
+        results = socket.getaddrinfo(hostname, 0):
```
juga0: Invalid syntax. There is a semicolon at the end of the line

PR #281, opened 2018-10-26
Ticket28061 priority

juga0:
Travis error comes from new flake8 version, not from changes in this branch


juga0:
i didn't see this comment before.
I got to think that using iterators would make the method being called for each iteration, (and i forgot to convert to list in my changes).
The *new thing* you propose might make sense, but yet, while that condition is not reached before the last iteration?.
And how is that i didn't get results close to each other this time?: https://trac.torproject.org/projects/tor/ticket/28061#comment:22

juga0:
I realized that change make sense. I made that change and left it running for a day. There're no relays measured within minutes.
There're still relays being measured within one hour, but that's other issue.
Thanks for the suggestion.
Rebased also to master.

PR #286, opened 2018-11-14
Bug28442: make sbws round to 3 significant figures in torflow rounding mode

teor2345:


juga0:
lgtm.
Do we then want to use 2 or 3 digits?. It's by default 3 now https://github.com/torproject/sbws/blob/821f86e8392bc05579f7785e6907be6cc3844fef/sbws/globals.py#L42

teor2345:
> Do we then want to use 2 or 3 digits?. It's by default 3 now

Let's start by matching torflow, which rounds to 3 significant digits.
I will open other tickets to move to 2 significant digits.

teor2345:
#28442 is closed.

PR #288, opened 2018-11-15
Bug28458

juga0:


juga0:
Need to fix CI...

sbws/core/scanner.py
```diff
@@ -177,10 +177,9 @@ def measure_relay(args, conf, destinations, cb, rl, relay):
     # exit, then pick a non-exit. Otherwise pick an exit.
     helper = None
     circ_fps = None
-    if relay.can_exit_to(dest.hostname, dest.port) and \
-            relay not in rl.bad_exits:
+    if relay not in rl.bad_exits:
```
pastly: BadExit is a flag that means an exit has been caught doing bad things (like injecting malicious JavaScript) and that clients should not use it for exiting. 99.9% of relays do not have this flag. This is not what you want to be checking here.

sbws/core/scanner.py
```diff
@@ -177,10 +177,9 @@ def measure_relay(args, conf, destinations, cb, rl, relay):
...
         helper = _pick_ideal_second_hop(
-            relay, dest, rl, cb.controller, is_exit=False)
+            relay, dest, rl, cb.controller, is_exit=True)
```
pastly: With this change, we are now picking a helper that must have the exit flag, and then we proceed to use it as the first hop in a circuit. This is not what you want to be doing here.

sbws/lib/destination.py
```diff
@@ -179,8 +179,7 @@ def _perform_usability_test(self):
         session = requests_utils.make_session(cont, timeout)
         usable_dests = []
         for dest in self._all_dests:
-            possible_exits = [e for e in self._rl.exits
-                              if e.can_exit_to(dest.hostname, dest.port)]
+            possible_exits = [e for e in self._rl.exits]
```
pastly: **If** it's a good idea to stop checking exit policies against the destination we want to use, then this should just be `possible_exits = self._rl.exits`.

You'll avoid errors caused by DNS failing (yay?) but will induce errors by trying to exit Tor from an exit that doesn't support our destination (which could have been prevented).

sbws/core/scanner.py
juga0: right, just realized that we ended up measuring in both ways

sbws/lib/destination.py
juga0: i'm actually implementing for checking if we're allow to exit without having to solve dns

sbws/core/scanner.py
juga0: i was trying to modify your code as minimum as possible

juga0:
I've solved what you comment locally, but before i push and you get the email notification (and then maybe you feel like you have to review it), see my comment in https://trac.torproject.org/projects/tor/ticket/22453#comment:32

juga0:
The new PR is https://github.com/torproject/sbws/pull/294

PR #290, opened 2018-11-15
Travis: use the high-availability keyserver pool

teor2345:
sbws is getting a significant number of build failures due to keyserver failures.

Using the high-availability keyserver pool reduces the probability of failure.


juga0:
@pastly: please remember that we stop merging from github months ago. We don't have automatic synchronization from gh to git.tpo, but we have the other way around.

PR #295, opened 2018-11-23
v3bwfile: stop using torflow_round_digs

juga0:
The generate cli argument changed name to round_digs but the
argument passed to V3BWFile.from_results was using the old
torflow_round_digs.

Closes bug #28602. Bugfix 1.0.3-dev0.

### DIFF
--- a/sbws/core/generate.py
+++ b/sbws/core/generate.py
@@ -2,7 +2,7 @@ from math import ceil
 
 from sbws.globals import (fail_hard, SBWS_SCALE_CONSTANT, TORFLOW_SCALING,
                           SBWS_SCALING, TORFLOW_BW_MARGIN, PROP276_ROUND_DIG,
-                          DAY_SECS, NUM_MIN_RESULTS, TORFLOW_ROUND_DIG)
+                          DAY_SECS, NUM_MIN_RESULTS)
 from sbws.lib.v3bwfile import V3BWFile
 from sbws.lib.resultdump import load_recent_results_in_datadir
 from argparse import ArgumentDefaultsHelpFormatter
@@ -54,13 +54,8 @@ def gen_parser(sub):
     p.add_argument('-m', '--torflow-bw-margin', default=TORFLOW_BW_MARGIN,
                    type=float,
                    help="Cap maximum bw when scaling as Torflow. ")
-    p.add_argument('-r', '--torflow-round-digs', default=TORFLOW_ROUND_DIG,
-                   type=int,
-                   help="Number of most significant digits to round "
-                        "bw in Torflow. This option is kept for compatibility "
-                        "with 1.0.x versions but it's silently ignored."
-                        "Use --round-digs instead.")
-    p.add_argument('-d', '--round-digs', default=PROP276_ROUND_DIG, type=int,
+    p.add_argument('-r', '--round-digs', '--torflow-round-digs',
+                   default=PROP276_ROUND_DIG, type=int,
                    help="Number of most significant digits to round bw.")
     p.add_argument('-p', '--secs-recent', default=None, type=int,
                    help="How many secs in the past are results being "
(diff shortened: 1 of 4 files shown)


sbws/core/generate.py
```diff
@@ -54,8 +54,13 @@ def gen_parser(sub):
     p.add_argument('-m', '--torflow-bw-margin', default=TORFLOW_BW_MARGIN,
                    type=float,
                    help="Cap maximum bw when scaling as Torflow. ")
-    p.add_argument('-r', '--round-digs', '--torflow-round-digs',
```
teor2345: When --torflow-round-digs is passed to sbws, sbws should do what it did in 1.0.

Please revert your changes to lines 57 and 58.

sbws/core/generate.py
```diff
@@ -2,7 +2,7 @@
 
 from sbws.globals import (fail_hard, SBWS_SCALE_CONSTANT, TORFLOW_SCALING,
                           SBWS_SCALING, TORFLOW_BW_MARGIN, PROP276_ROUND_DIG,
-                          DAY_SECS, NUM_MIN_RESULTS)
+                          DAY_SECS, NUM_MIN_RESULTS, TORFLOW_ROUND_DIG)
```
teor2345: When you revert 57 and 58, you will also need to revert this change.

sbws/core/generate.py
```diff
@@ -106,7 +111,7 @@ def main(args, conf):
     bw_file = V3BWFile.from_results(results, state_fpath, args.scale_constant,
                                     scaling_method,
                                     torflow_cap=args.torflow_bw_margin,
-                                    torflow_round_digs=args.torflow_round_digs,
+                                    round_digs=args.round_digs,
```
teor2345: This change should work, because --round-digs is the first long option argument in the list:
https://docs.python.org/3/library/argparse.html#dest

teor2345: This change should work without any changes to the arguments.

juga0: Yes, but then you suggest but then we're never using --torflow-round-digs and we can't cause it won't be in the args Namespace.
You said in a previous comment:
> When --torflow-round-digs is passed to sbws, sbws should do what it did in 1.0.

teor2345: I don't understand what the problem is here.

If you revert new lines 57-63 to old lines 57-58 in master, then:
1. args.round_digs should exist:
https://docs.python.org/3/library/argparse.html#dest
2. when --torflow-round-digs is passed to sbws, sbws will do what it did in 1.0

Can you show me a test where this does not work?

juga0: > I don't understand what the problem is here.
> 
> If you revert new lines 57-63 to old lines 57-58 in master, then:
> 
>     1. args.round_digs should exist:
>        https://docs.python.org/3/library/argparse.html#dest
correct
>     2. when --torflow-round-digs is passed to sbws, sbws will do what it did in 1.0

no, --torflow-round-digs won't be in the namespace and it's not passed in V3BWFile.from_file in your suggestion.
> 
> Can you show me a test where this does not work?

I can create a test, but if you just checkout current master and run 'sbws generate`, you get:
`AttributeError: 'Namespace' object has no attribute 'torflow_round_digs'`
If you try `sbws generate --torflow-round-digs`, you get:
`sbws generate: error: argument -r/--round-digs/--torflow-round-digs: expected one argument`

juga0: Oh, sorry, you were right, got confused here.
I couldn't revert just those lines because would lose other changes.
I created a fixup that reverts those lines, and use torflow's rounding method in the test.
And added a new test for those options.

juga0:
squashed and merged.
Because  tests/unit/lib/test_v3bwfile.py was conflicting and was modified, GH didn't detect the branch merge.

PR #306, opened 2018-12-04
Make sbws actually read user-specified config if given

pastly:


sbws/util/config.py
```diff
@@ -70,7 +70,8 @@ def _get_user_config(args, conf=None):
             # sbws should start with a logger before reading configurations.
             print('Configuration file %s not found, using defaults.' %
                   args.config)
-        return _extend_config(conf, args.config)
```
juga0: I think it would be more clear if you return conf in the if
```suggestion
            return conf
```

sbws/util/config.py
```diff
@@ -70,7 +70,8 @@ def _get_user_config(args, conf=None):
...
+        else:
```
juga0: i think it would be more clear if instead of else you return _extend_config here
```suggestion
        return _extend_config(conf, args.config)
```

juga0:
Two commits to change to lines seems excesive to me

juga0:
There's a test not finished, and would be great to add more tests for all the cases

sbws/util/config.py
pastly: It would also cause us to try to read a file that doesn't exist.

juga0: No when the return conf in https://github.com/torproject/sbws/pull/306#discussion_r239874182 is applied first

juga0:
Closing in favor of https://github.com/torproject/sbws/pull/326

PR #316, opened 2018-12-15
Bug28692

juga0:


sbws/util/stem.py
```diff
@@ -117,6 +122,33 @@ def _init_controller_socket(socket):
     return c
 
 
+def set_torrc_runtime_options(cont):
+    """Set torrc options at runtime."""
+    try:
+        cont.set_conf('__DisablePredictedCircuits', '1')
+        cont.set_conf('__LeaveStreamsUnattached', '1')
```
teor2345: Can you refactor these options into a dictionary constant?
And let's put all the options in the same place.
You could put the new constant in globals.py after TORRC_STARTING_POINT.

sbws/util/stem.py
```diff
@@ -16,6 +18,9 @@
 stream_building_lock = RLock()
 
 
+TORRC_OPTIONS_CAN_FAIL = OrderedDict({'ConnectionPadding': '0'})
```
teor2345: Let's put all the options in the same place.
You could put this constant in globals.py after TORRC_STARTING_POINT.

sbws/util/stem.py
```diff
@@ -117,6 +122,33 @@ def _init_controller_socket(socket):
     return c
 
 
+def set_torrc_runtime_options(cont):
```
teor2345: Please call this function on line 74 as well, to remove duplicate code.

sbws/util/stem.py
juga0: That was my initial idea (https://trac.torproject.org/projects/tor/ticket/28692#comment:8):
```
>  Even if only stem.py uses these options, i think they should be in globals.py, since we might need to change them and globals.py should be the place where to change sbws defaults.

Let's leave any refactoring until #28737.
```

sbws/util/stem.py
juga0: That was my initial idea (https://trac.torproject.org/projects/tor/ticket/28692#comment:8)

sbws/util/stem.py
juga0: yes, i forgot

sbws/util/stem.py
```diff
@@ -117,6 +117,32 @@ def _init_controller_socket(socket):
...
+        log.exception(e)
+        exit(1)
+
+
+def set_torrc_options_can_fail(controller):
+    """Set options that can fail, at runtime.
+    They can be set at launch, but since the may fail because they are not
+    supported in some Tor versions, it's easier to try one by one at runtime
+    and ignore the ones that fail.
+
```
teor2345: Typo: extra newline?

sbws/util/stem.py
```diff
@@ -72,8 +73,7 @@ def init_controller(port=None, path=None, set_custom_stream_settings=True):
             return None, 'Unable to reach tor on control socket'
     assert c is not None
     if set_custom_stream_settings:
-        c.set_conf('__DisablePredictedCircuits', '1')
-        c.set_conf('__LeaveStreamsUnattached', '1')
+        set_torrc_runtime_options(c)
```
teor2345: Do we also need to call `set_torrc_options_can_fail(c)` here?

sbws/globals.py
```diff
@@ -26,6 +29,13 @@
     'ProtocolWarnings': '1',
 }
 
+TORRC_RUNTIME_OPTIONS = {
```
teor2345: Do we need a comment here?

sbws/globals.py
```diff
@@ -26,6 +29,13 @@
...
+    '__DisablePredictedCircuits': '1',
+    '__LeaveStreamsUnattached': '1',
+}
+
+TORRC_OPTIONS_CAN_FAIL = OrderedDict({'ConnectionPadding': '0'})
```
teor2345: Do we need a comment here that explains how we use these options?

juga0:
squashed and rebased in https://github.com/juga0/sbws/commits/bug28692_squashed_rebased.

juga0:
Merged the squashed branch.

PR #320, opened 2018-12-22
Bug28897

juga0:


sbws/lib/destination.py
```diff
@@ -76,21 +75,26 @@ def connect_to_destination_over_circuit(dest, circ_id, session, cont, max_dl):
             head = requests_utils.head(session, dest.url, verify=dest.verify)
         except (requests.exceptions.ConnectionError,
                 requests.exceptions.ReadTimeout) as e:
+            dest.set_failure
```
teor2345: Using a property to change state makes this code hard to read.
Should it be a function call instead?
This line is repeated 4 times in this function: please fix all of them.

sbws/lib/destination.py
```diff
@@ -76,21 +75,26 @@ def connect_to_destination_over_circuit(dest, circ_id, session, cont, max_dl):
...
         return False, error_prefix + 'we except the header Content-Length '\
-                'to exist in the response'
+            'to exist in the response'
     content_length = int(head.headers['content-length'])
     if max_dl > content_length:
+        dest.set_failure
         return False, error_prefix + 'our maximum configured download size '\
             'is {} but the content is only {}'.format(max_dl, content_length)
     log.debug('Connected to %s over circuit %s', dest.url, circ_id)
+    dest.failed = False
```
teor2345: Where is dest.failed set to True?
Should we set it to True at the start of this function?

sbws/lib/destination.py
```diff
@@ -103,17 +107,39 @@ def __init__(self, url, max_dl, verify):
         assert u.netloc
         self._url = u
         self._verify = verify
+        # Flag to record whether this destination failed in the last
+        # measurement.
+        # Failures can't happen if:
```
teor2345: Do you mean "can happen"?

sbws/globals.py
```diff
@@ -53,6 +53,10 @@
 
 BW_LINE_SIZE = 510
 
+# This number might need adjusted depending on the percentage of circuits and
+# HTTP requests failures.
+MAXIMUM_NUMBER_DESTINATION_FAILURES = 10
```
teor2345: Will operators want to configure this value?

sbws/core/scanner.py
```diff
@@ -169,10 +170,11 @@ def measure_relay(args, conf, destinations, cb, rl, relay):
         cb.controller, conf.getfloat('general', 'http_timeout'))
     # Pick a destionation
     dest = destinations.next()
+    # If there is no any destination at this point, it can not continue.
     if not dest:
-        log.warning('Unable to get destination to measure %s %s',
-                    relay.nickname, relay.fingerprint[0:8])
-        return None
+        log.critical("There are not any functional destinations.")
```
teor2345: Should we tell operators what to do to avoid this error?

For example:
"If you add multiple destinations to your config, sbws will use all the working destinations."

sbws/lib/destination.py
```diff
@@ -245,23 +219,10 @@ def next(self):
...
-
-        self._rng.shuffle(self._usable_dests)
-        return self._usable_dests[0]
+        # Do not perform usability tests since a destination is already proven
+        # usable or not in every measurement, and it should depend on a X
+        # number of failures.
+        # This removes the need for an extra lock for every measurement.
+        # Do not change the order of the destinations, just return a
+        # destination.
+        return self._rng.choice(self.functional_destinations)
```
teor2345: It looks like sbws will slowly reduce its list of destinations on every network error, until there are none left.
Can a destination ever recover after it is excluded?

There are two cases that we need to handle:
* If a network error is only in one destination, and it lasts for a short time, sbws should use the destination once the network starts working again.
* If a network error affects all destinations, and it lasts for a long time, sbws should error.

Can we retry non-functional destinations with a low probability?

Can we do a few slow retries before failing?
Or can we wait for a few minutes before failing?

sbws/core/scanner.py
```diff
@@ -169,10 +170,11 @@ def measure_relay(args, conf, destinations, cb, rl, relay):
...
+        # This should raise an error so that the caller can close the pool.
+        exit(1)
```
teor2345: What happens when all the pools fail?

sbws/core/scanner.py
teor2345: Should we tell operators that multiple destinations are more reliable in the config or the documentation?

juga0:
> This code looks good.
> 
> Here are the design changes I think it might need:
> 
>     * sbws should be able to recover if the network fails for a short time, then starts working again. Recovery should work if one destination fails, and if all destinations fail.
> 
>     * operators might want to configure MAXIMUM_NUMBER_DESTINATION_FAILURES
> 
> 
> What do you think?

I think that we should open other ticket to implement that in other milestone.
It's true that if there're network problems right now sbws can not recover, but at least it's an easy problem to identify in other ways.

If you agree, then i'd change the other issues.

teor2345:
> I think that we should open other ticket to implement that in other milestone.
> It's true that if there're network problems right now sbws can not recover, but at least it's an easy problem to identify in other ways.

Ok, can you open a ticket in sbws 1.1 to make destinations recover?

Since destinations will fail forever in 1.0, let's make the default MAXIMUM_NUMBER_DESTINATION_FAILURES equal to 100.

sbws/lib/destination.py
juga0: Yes, fixed

sbws/lib/destination.py
juga0: Yes, fixed.

sbws/core/scanner.py
juga0: Added comment to the config and doc, since dunno where operators look first.

sbws/lib/destination.py
juga0: Created #29589

sbws/globals.py
juga0: Created #29589

sbws/core/scanner.py
juga0: There's only one pool. The caller is the callback, i though it's better to just call  `stop_threads` (merged in master) from here (https://github.com/juga0/sbws/commit/3fe64b06253f9342e627c5d4b9db065cd03a8a7e, in https://github.com/juga0/sbws/tree/bug28897_squashed_rebased)

sbws/lib/destination.py
juga0: It can not be set to True at the start, because then we would lose the information about if it failed before. I added a comment in the code.

juga0:
> > I think that we should open other ticket to implement that in other milestone.
> > It's true that if there're network problems right now sbws can not recover, but at least it's an easy problem to identify in other ways.
> 
> Ok, can you open a ticket in sbws 1.1 to make destinations recover?

#29589

> Since destinations will fail forever in 1.0, let's make the default MAXIMUM_NUMBER_DESTINATION_FAILURES equal to 100.

ok, done.

juga0:
> Can you please do a rebase or merge before the next review, so I can review the code that will be merged to master?

https://github.com/torproject/sbws/pull/339

We can close this PR after you review fixups.

teor2345:
The fixups seem fine to me. Closing because #339 is newer.

PR #322, opened 2019-01-05
Stop including tests network as binary blob

juga0:
To easy diffs and avoid security issues.

Fixes bug #28590. Bugfix v0.4.0.


juga0:
Rebased to master in https://github.com/juga0/sbws/tree/bug28590_rebased_master because travis fails to fetch the key from key servers.

PR #324, opened 2019-01-08
Bug28869

juga0:


sbws/__init__.py
```diff
@@ -1 +1,20 @@
 __version__ = '1.0.3-dev0'
+
+
+import threading
+
+from . import globals  # noqa
+
+
```
asn-d6: Hmm, I don't quite understand what's going on here and the commit does not explain what this is used for?
Perhaps add some more docs to explain this addition to the init file? What are we trying to achieve?

sbws/core/scanner.py
```diff
@@ -270,11 +270,7 @@ def measure_relay(args, conf, destinations, cb, rl, relay):
 
 
 def dispatch_worker_thread(*a, **kw):
-    try:
-        return measure_relay(*a, **kw)
-    except Exception as err:
-        log.exception('Unhandled exception in worker thread')
-        raise err
+    return measure_relay(*a, **kw)
```
asn-d6: Does this stop us from raising an exception? Doesn't it still raise an exception if it triggers?

sbws/core/scanner.py
```diff
@@ -119,7 +118,7 @@ def measure_bandwidth_to_server(session, conf, dest, content_length):
         'target': conf.getfloat('scanner', 'download_target'),
         'max': conf.getfloat('scanner', 'download_max'),
     }
-    while len(results) < num_downloads:
+    while len(results) < num_downloads and not settings.end_event.is_set():
```
asn-d6: Hmm, why are we checking the `end_event` here? This does not seem to be part of the `scanner: use end event singleton` commit message explanation.

sbws/core/scanner.py
```diff
@@ -354,6 +354,37 @@ def closure(object):
...
 
 
+def main_loop(args, conf, controller, relay_list, circuit_builder, result_dump,
+              relay_prioritizer, destinations, max_pending_results, pool):
+    pending_results = []
+    while not settings.end_event.is_set():
+        num_relays = 0
+        loop_tstart = time.time()
+        for target in relay_prioritizer.best_priority():
+            if settings.end_event.is_set():
```
asn-d6: Why do we need to check settings.end_event.is_set() twice in this function?

sbws/core/scanner.py
```diff
@@ -354,6 +354,37 @@ def closure(object):
...
+            callback = result_putter(result_dump)
+            callback_err = result_putter_error(target)
+            async_result = pool.apply_async(
+                dispatch_worker_thread,
+                [args, conf, destinations, circuit_builder, relay_list,
+                 target], {}, callback, callback_err)
+            pending_results.append(async_result)
+            log.debug(async_result)
+            while len(pending_results) >= max_pending_results:
+                time.sleep(5)
```
asn-d6: This is not part of this new patch, but what's the point of sleeping for 5 seconds?

sbws/core/scanner.py
juga0: Yes, an exception will be raised if it's triggered. Wrong commit message, i just wanted to remove the log.

sbws/core/scanner.py
juga0: hmm, it does use the ``end_event`` singleton.
The intention is to don't start new request when sbws is ending.

juga0: Otherwise, finishing sbws would mean to wait for all the new requests.

sbws/core/scanner.py
juga0: To don't start a new loop measuring relays (it takes ~3 hours) when sbws is finishing.

sbws/core/scanner.py
juga0: Every `async_result` has its own thread and the thread are reused. I assumed that was needed to don't reuse a thread before the `async_result` is finished, but maybe it's not need.

sbws/core/scanner.py
asn-d6: ACK

sbws/core/scanner.py
asn-d6: ACK!

sbws/core/scanner.py
asn-d6: Hmm, I'm not sure if sleeping is the right way to prevent anything in multi-threading code but I also don't understand exactly what is meant to be achieved. Perhaps adding a comment about it, might help us remember it later if we want to discuss it in Brussels?

Also, should we add function documentation to this new main_loop function?

sbws/__init__.py
```diff
@@ -1,17 +1,22 @@
...
-
-from . import globals  # noqa
+import threading  # noqa
 
 
 class Settings:
+    """Singleton settings for all the packages.
+    This way change settings can be seen by all the packages that import it.
+
+    It leaves in ``__init__.py`` to leave open the possibility of having a
```
asn-d6: Thanks for the comment.

Nitpick: s/It leaves/It lives/

sbws/core/scanner.py
juga0: Added in e4e671d. Maybe your questions are now answered with the documentation?.

juga0:
Squashed and merged.

PR #328, opened 2019-01-10
Bug28868

juga0:


sbws/lib/relayprioritizer.py
```diff
@@ -23,71 +22,100 @@ def __init__(self, args, conf, relay_list, result_dump):
...
-                    # that they are not de-prioritized as much. This way, we
-                    # will come back to them sooner to try again.
-                    assert result.freshness_reduction_factor >= 0.0
-                    assert result.freshness_reduction_factor <= 1.0
-                    log.debug('Cutting freshness for a %s result by %d%% for '
-                              '%s', result.type.value,
-                              result.freshness_reduction_factor * 100,
-                              relay.nickname)
+                if isinstance(result, ResultError) \
+                        and prioritize_result_error is True:
```
ahf: I think this one can just be `prioritize_result_error` and you can leave out the `is True`?

juga0: good point, i think that to avoid errors caused by python not being statically typed, is better to check the type when there's ambiguity.
For instance, if at some point we want to prioritize relays that have more than 3 errors (and we reuse the same argument or in the same position because of a human mistake), and we pass to the method `prioritize_result_error=2`, the code would be executed, when it's not what was intended. Makes sense?

PR #332, opened 2019-02-06
Ticket29294

juga0:


scripts/maint/release.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+"""Script to help release new versions."""
+import sys
+
+
+def main(args):
+    version = args[0]
+    print("# This script will guide you to create a new release, but it will "
+          "not do anything.")
```
teor2345: I don't understand why this is a script?
Can it just be a document with instructions?

CONTRIBUTING.rst
```diff
@@ -136,6 +136,9 @@ Commits
 Each commit should reference the Tor Project Trac ticket (example: ``#12345``)
 and possibly the bugfix version.
 
+From version 1.0.3, the commit title should start by ``new`` or ``fix`` or
+``change``. The commit message should contain ``Closes: #bugnumber``.
```
teor2345: A git commit hook could help you enforce this rule.
But I think that should be a different ticket.

scripts/maint/release.py
juga0: Because it should actually create the new release at some point.
For now just prints the correct versions, a bit more help that just instructions.

CONTRIBUTING.rst
juga0: ok, created #29582

scripts/maint/release.py
```diff
@@ -1,42 +1,171 @@
...
+          " correct tag...")
+    input("Press enter when you are sure everything is correct.")
+    subprocess.call(['git', 'push', 'origin', 'master'])
+
+    # While we use Github releases and not dist.tpo
+    print("\n6. Create release tarball")
+    print("-------------------------")
+    print("Obtaining Github tarball...")
+    subprocess.call(
+        "torsocks wget https://github.com/torproject/sbws/archive/v{}.tar.gz"
```
teor2345: Will everyone have torsocks?
Does GitHub support tor?
(Do some exits fail on GitHub?)

Maybe the user can call the script like "torsocks python scripts/maint/release.py" if they want to use torsocks?

scripts/maint/release.py
```diff
@@ -1,42 +1,171 @@
...
+        .format(release_version).split(' ')
+        )
+
+    print("\n7. Create the tarball signature")
+    print("-------------------------------")
+    print("Creating detached signature...")
+    subprocess.call("gpg --default-key F305447AF806D46B "
```
teor2345: Whose key is this?
It should probably be an argument to the script.

scripts/maint/release.py
juga0: Github supports tor, didn't fail for me in years.
Better, just check that the remotely obtained tarball hash is the same as one generated locally, what also ensures that we're in the correct branch/dir.

scripts/maint/release.py
juga0: that's juga's key. Agree, just didn't do it cause forgot.

scripts/maint/release.py
```diff
@@ -132,12 +132,36 @@ def main(args):
...
+        .format(release_version, release_version, release_version).split(' ')
+    )
+    print("\nCreating tarball hash file...")
+    fd = open('v{}.tar.gz.sha256'.format(release_version), 'w')
+    subprocess.call("sha256sum v{}.tar.gz".format(release_version).split(' '),
+                    stdout=fd)
+    fd.close()
+
     print("Obtaining Github tarball...")
+    # This will overwrite local tarball, but that's fine since the hash file
```
teor2345: Overwriting the local tarball is only fine if the user reads the output of the script, and notices when the GitHub tarball is different. Please put the tarballs in separate files.

scripts/maint/release.py
```diff
@@ -132,12 +132,36 @@ def main(args):
...
+    # won't be overwritten.
     subprocess.call(
-        "torsocks wget https://github.com/torproject/sbws/archive/v{}.tar.gz"
-        .format(release_version).split(' ')
+        "wget https://github.com/torproject/sbws/archive/v{}.tar.gz "
+        "-O v{}.tar.gz"
+        .format(release_version, release_version).split(' ')
         )
 
+    print("Verifying Github tarball and local one are the same...")
```
teor2345: I don't understand why we do a sha256 check, rather than a file diff check.
In fact, I'm not sure why we do this check at all?
Please add a comment explaining why it is important that GitHub matches our local tarball.

Are tarballs reproducible?
What happens if the tarballs are different?
Can they be different on different OSes?
Does GitHub guarantee that their tarballs are created with particular git and tar versions on a particular OS?

PR #338, opened 2019-02-26
Ticket29299 squashed rebased

juga0:


docs/source/man_sbws.ini.rst
```diff
@@ -72,6 +72,9 @@ destinations.STR
   verify = BOOL
     Whether or not to verify the destination certificate.
     (Default: True)
+  country = STR
+    ISO 3166-1 alpha-2 country code.
+    Use 00 if the destination URL is a domain name and it is in a CDN.
```
tlyu: Maybe use another alphabetic user-assigned code like `ZZ`? It seems odd to use numeric `00` here when there are still some user-assigned codes available.

sbws/core/scanner.py
```diff
@@ -557,7 +557,6 @@ def gen_parser(sub):
 def main(args, conf):
     if conf.getint('scanner', 'measurement_threads') < 1:
         fail_hard('Number of measurement threads must be larger than 1')
-
```
tlyu: Please don't include formatting changes in a commit that contains code changes. If this reformatting needs to happen, it should go in a separate commit.

docs/source/man_sbws.ini.rst
juga0: Changed to ZZ

sbws/core/scanner.py
juga0: Was a mistake. Fixed

sbws/core/scanner.py
```diff
@@ -571,5 +571,4 @@ def main(args, conf):
     # Generate an unique identifier for each scanner
     if 'uuid' not in state:
         state['uuid'] = str(uuid.uuid4())
-
```
tlyu: This is one more extraneous minor format change.

juga0: ok, fixup to solve that

juga0:
Squashed and force pushed

PR #345, opened 2019-03-07
Ticket28566 rebased on 29591

juga0:


sbws/lib/resultdump.py
```diff
@@ -215,15 +216,18 @@ def __init__(self, fingerprint, nickname, address, master_key_ed25519,
             self.consensus_bandwidth = consensus_bandwidth
             self.consensus_bandwidth_is_unmeasured = \
                 consensus_bandwidth_is_unmeasured
+            self.consensus_count = consensus_count
```
nmathewson: I think this new field should be documented somewhere.  You should probably add documentation, unless some later commit already does that.

sbws/lib/relaylist.py
```diff
@@ -173,6 +173,10 @@ def update_consensus_timestamps(self, timestamp=None):
         self._set_new_consensus_timestamps(timestamp)
         self._remove_old_consensus_timestamps
 
+    @property
+    def consensus_count(self):
```
nmathewson: Needs documentation?

sbws/lib/relaylist.py
```diff
@@ -331,6 +335,10 @@ def _refresh(self):
         # Set a new list of relays.
         self._relays = self._init_relays()
 
+    @property
+    def consensus_count(self):
```
nmathewson: Needs documentation?

sbws/lib/relaylist.py
```diff
@@ -1,16 +1,38 @@
...
-import time
 import logging
 from threading import Lock
 
+from ..globals import MEASUREMENTS_PERIOD
+
 log = logging.getLogger(__name__)
 
 
+def remove_old_consensus_timestamps(
```
nmathewson: documentation

sbws/lib/relaylist.py
```diff
@@ -1,16 +1,38 @@
...
+        consensus_timestamps, measurements_period=MEASUREMENTS_PERIOD):
+    oldest_date = datetime.utcnow() - timedelta(measurements_period)
+    new_consensus_timestamps = \
+        [t for t in consensus_timestamps if t >= oldest_date]
+    return new_consensus_timestamps
+
+
+def valid_after_from_network_statuses(network_statuses):
```
nmathewson: needs documentation

sbws/lib/relaylist.py
```diff
@@ -1,16 +1,38 @@
...
+        if document:
+            valid_after = getattr(document, 'valid_after', None)
+            if valid_after:
+                return valid_after
+    return datetime.utcnow().replace(microsecond=0)
+
+
 class Relay:
-    def __init__(self, fp, cont, ns=None, desc=None):
+    def __init__(self, fp, cont, ns=None, desc=None, timestamp=None):
```
nmathewson: Should document what the timestamp is a timestamp for.

sbws/lib/relaylist.py
```diff
@@ -107,6 +131,48 @@ def master_key_ed25519(self):
             return None
         return key.rstrip('=')
 
+    @property
+    def consensus_valid_after(self):
```
nmathewson: documentation

sbws/lib/relaylist.py
```diff
@@ -107,6 +131,48 @@ def master_key_ed25519(self):
...
+        network_status_document = self._from_ns('document')
+        if network_status_document:
+            return getattr(network_status_document, 'valid_after', None)
+        return None
+
+    @property
+    def last_consensus(self):
```
nmathewson: documentation

nmathewson: Should this be called last_consensus_timestamp or last_consensus_valid_after?  With the current name, it sounds like it returns a consensus rather than a timestamp.

sbws/lib/relaylist.py
```diff
@@ -107,6 +131,48 @@ def master_key_ed25519(self):
...
+        if len(self._consensus_timestamps) >= 1:
+            return self._consensus_timestamps[-1]
+        return None
+
+    def _set_new_consensus_timestamps(self, timestamp=None):
```
nmathewson: Should this be called "add" instead of "set?"  Needs documentation.

sbws/lib/relaylist.py
```diff
@@ -107,6 +131,48 @@ def master_key_ed25519(self):
...
+                self._consensus_timestamps.append(self.consensus_valid_after)
+        elif (self.last_consensus is not None
+              and timestamp > self.last_consensus):
+            self._consensus_timestamps.append(timestamp)
+        else:
+            self._consensus_timestamps.append(
+                datetime.utcnow().replace(microsecond=0))
+
+    @property
+    def _remove_old_consensus_timestamps(
```
nmathewson: This doesn't seem like a property me; it looks a lot more like it should be a function.

sbws/lib/relaylist.py
```diff
@@ -129,16 +195,38 @@ class RelayList:
...
+        # To track all the consensus seen.
+        self._consensus_timestamps = []
+        # Initialize so that there's no error trying to access to it.
+        # In future refactor, change to a dictionary, where the keys are
+        # the relays' fingerprint.
+        self._relays = []
+        self._measurements_period = measurements_period
         self._refresh()
 
     def _need_refresh(self):
```
nmathewson: documentation

sbws/lib/relaylist.py
```diff
@@ -129,16 +195,38 @@ class RelayList:
...
-        return time.time() >= self._last_refresh + self.REFRESH_INTERVAL
+        # New consensuses happen every hour.
+        return datetime.utcnow() >= \
+            self.last_consensus + timedelta(seconds=60*60)
+
+    @property
+    def last_consensus(self):
```
nmathewson: I suggest naming this last_consensus_received_at or something like that -- this property sounds like it should return the last consensus itself.

sbws/lib/relaylist.py
```diff
@@ -197,19 +285,51 @@ def _relays_with_flag(self, flag):
     def _relays_without_flag(self, flag):
         return [r for r in self.relays if flag not in r.flags]
 
+    @property
+    def _remove_old_consensus_timestamps(self):
```
nmathewson: Again, I don't think this makes sense as a property.

juga0:
Agree with all suggestions.
Squashed fixups and force pushed since not expecting new review

PR #347, opened 2019-03-11
Ticket28567 rebased

juga0:


teor2345:
It looks like this was squashed and merged as e8f7d9d.

PR #348, opened 2019-03-12
Ticket28565 rebased

juga0:
To review only from https://github.com/torproject/sbws/commit/70531ba0fa0e424480d8b9b91f2bfac5618e972e included

### DIFF
--- a/sbws/lib/v3bwfile.py
+++ b/sbws/lib/v3bwfile.py
@@ -525,20 +525,51 @@ class V3BWLine(object):
                 str(max(relay_recent_priority_list_counts))
 
         success_results = [r for r in results if isinstance(r, ResultSuccess)]
+
+        # NOTE: The following 4 conditions exclude relays from the bandwidth
+        # file when the measurements does not satisfy some rules, what makes
+        # the relay non-`eligible`.
+        # In BANDWIDTH_LINE_KEY_VALUES_MONITOR it is explained what they mean.
+        # In BW_HEADER_KEYVALUES_RECENT_MEASUREMENTS_EXCLUDED it is also
+        # explained the what it means the strings returned.
+        # They rules were introduced in #28061 and #27338
+        # In #28565 we introduce the KeyValues to know why they're excluded.
+        # In #28563 we report these relays, but make Tor ignore them.
+        # This might confirm #28042.
+
+        number_excluded_error = len(results) - len(success_results)
+        if number_excluded_error > 0:
+            # then the number of error results is the number of results
+            kwargs['relay_recent_measurements_excluded_error_count'] = \
+                number_excluded_error
         if not success_results:
-            return None
+            return None, 'recent_measurements_excluded_error_count'
+
         results_away = \
             cls.results_away_each_other(success_results, secs_away)
+        number_excluded_near = len(success_results) - len(results_away)
+        if number_excluded_near > 0:
+            kwargs['relay_recent_measurements_excluded_near_count'] = \
+                len(success_results) - len(results_away)
         if not results_away:
-            return None
+            return None, 'recent_measurements_excluded_near_count'
+
         # log.debug("Results away from each other: %s",
         #           [unixts_to_isodt_str(r.time) for r in results_away])
         results_recent = cls.results_recent_than(results_away, secs_recent)
+        number_excluded_old = len(results_away) - len(results_recent)
+        if number_excluded_old > 0:
+            kwargs['relay_recent_measurements_excluded_old_count'] = \
+                number_excluded_old
         if not results_recent:
-            return None
+            return None, 'recent_measurements_excluded_old_count'
+
         if not len(results_recent) >= min_num:
+            kwargs['relay_recent_measurements_excluded_few_count'] = \
+                len(results_recent)
             # log.debug('The number of results is less than %s', min_num)
-            return None
+            return None, 'recent_measurements_excluded_few_count'
+
         rtt = cls.rtt_from_results(results_recent)
         if rtt:
             kwargs['rtt'] = rtt
@@ -560,7 +591,7 @@ class V3BWLine(object):
         kwargs['desc_bw_obs_mean'] = \
             cls.desc_bw_obs_mean_from_results(results_recent)
         bwl = cls(node_id, bw, **kwargs)
-        return bwl
+        return bwl, None
 
     @classmethod
     def from_data(cls, data, fingerprint):
@@ -594,7 +625,7 @@ class V3BWLine(object):
                 return results
         # log.debug("Results are NOT away from each other in at least %ss: %s",
         #           secs_away, [unixts_to_isodt_str(r.time) for r in results])
-        return None
+        return []
 
     @staticmethod
     def results_recent_than(results, secs_recent=None):
@@ -769,8 +800,8 @@ class V3BWFile(object):
         state = State(state_fpath)
         for fp, values in results.items():
             # log.debug("Relay fp %s", fp)
-            line = V3BWLine.from_results(values, secs_recent, secs_away,
-                                         min_num)
+            line, reason = V3BWLine.from_results(values, secs_recent,
+                                                 secs_away, min_num)
             if line is not None:
                 bw_lines_raw.append(line)
         if not bw_lines_raw:
--- a/sbws/lib/v3bwfile.py
+++ b/sbws/lib/v3bwfile.py
@@ -453,6 +453,15 @@ class V3BWHeader(object):
         [setattr(self, k, str(v)) for k, v in kwargs.items()
          if k in STATS_KEYVALUES]
 
+    def add_relays_excluded_counters(self, exclusion_dict):
+        """
+        Add the monitoring KeyValues to the header about the number of
+        relays not included because they were not ``eligible``.
+        """
+        log.debug("Adding relays excluded counters.")
+        for k, v in exclusion_dict.items():
+            setattr(self, k, str(v))
+
 
 class V3BWLine(object):
     """
@@ -798,12 +807,26 @@ class V3BWFile(object):
         number_consensus_relays = cls.read_number_consensus_relays(
             consensus_path)
         state = State(state_fpath)
+
+        # Create a dictionary with the number of relays excluded by any of the
+        # of the filtering rules that makes relays non-`eligible`.
+        # NOTE: In BW_HEADER_KEYVALUES_RECENT_MEASUREMENTS_EXCLUDED it is
+        # explained what are the KeyValues.
+        # See also the comments in `from_results`.
+        exclusion_dict = dict(
+            [(k, 0) for k in BW_HEADER_KEYVALUES_RECENT_MEASUREMENTS_EXCLUDED]
+            )
         for fp, values in results.items():
             # log.debug("Relay fp %s", fp)
             line, reason = V3BWLine.from_results(values, secs_recent,
                                                  secs_away, min_num)
             if line is not None:
                 bw_lines_raw.append(line)
+            else:
+                exclusion_dict[reason] = exclusion_dict.get(reason, 0) + 1
+        # Add the headers with the number of excluded relays by reason
+        header.add_relays_excluded_counters(exclusion_dict)
+
         if not bw_lines_raw:
             log.info("After applying restrictions to the raw results, "
                      "there is not any. Scaling can not be applied.")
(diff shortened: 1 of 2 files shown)


sbws/lib/v3bwfile.py
```diff
@@ -528,6 +528,18 @@ def from_results(cls, results, secs_recent=None, secs_away=None,
...
+        # NOTE: The following 4 conditions exclude relays from the bandwidth
+        # file when the measurements does not satisfy some rules, what makes
+        # the relay non-`eligible`.
+        # In BANDWIDTH_LINE_KEY_VALUES_MONITOR it is explained what they mean.
+        # In BW_HEADER_KEYVALUES_RECENT_MEASUREMENTS_EXCLUDED it is also
+        # explained the what it means the strings returned.
+        # They rules were introduced in #28061 and #27338
+        # In #28565 we introduce the KeyValues to know why they're excluded.
+        # In #28563 will include again these lines, but make them unparseable
+        # by Tor.
```
teor2345: This comment is confusing: it sounds like you will make the file unparseable by Tor, which is bad. What you actually want is to make Tor ignore some of the relay lines in the file.

sbws/lib/v3bwfile.py
```diff
@@ -43,6 +43,18 @@
...
                    'minimum_percent_eligible_relays']
+BW_HEADER_KEYVALUES_RECENT_MEASUREMENTS_EXCLUDED = [
+    # The number of success results should be:
+    # the number of attempts - the number of failures
+    # 4.6 header: the number of successful results, created in the last 5 days,
+    # that were excluded by a filter
+    # This is the sum of the following 3 + not success results
+    # 'recent_measurement_exclusion_count',
+    'recent_measurement_exclusion_not_success_count',
+    'recent_measurement_exclusion_not_distanciated_count',
```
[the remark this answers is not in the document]
juga0: how does `distant` sounds?, better `away`?

sbws/lib/v3bwfile.py
```diff
@@ -43,6 +43,25 @@
...
                    'number_consensus_relays', 'percent_eligible_relays',
                    'minimum_percent_eligible_relays']
+BW_HEADER_KEYVALUES_RECENT_MEASUREMENTS_EXCLUDED = [
+    # The number of success results should be:
+    # the number of attempts - the number of failures
+    # 4.6 header: the number of successful results, created in the last 5 days,
+    # that were excluded by a filter
+    # This is the sum of the following 3 + not success results
+    # 'recent_measurement_exclusion_count',
+    'recent_measurement_exclusion_not_success_count',
```
teor2345: This key name is hard to understand, because it uses two negatives in a row: "exclusion not success".

Usually, we use a short name for the total, and then add extra information for each part of the total:
```suggestion
    'recent_measurement_exclusion_count',
```

sbws/lib/v3bwfile.py
```diff
@@ -43,6 +43,25 @@
...
+    # The relay measurements are not done in X time away from each other.
+    # By default this is 1 day.
+    'recent_measurement_exclusion_not_distanciated_count',
```
teor2345: Don't use two negatives "exclusion not distant", it's confusing.

Instead, explain what is being counted:

These recent measurements are in a short time range. Here is the exclusion count for those measurements.
```suggestion
    'recent_measurement_in_short_time_range_exclusion_count',
```

sbws/lib/v3bwfile.py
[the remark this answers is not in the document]
teor2345: Distant is better. But don't use two negatives "exclusion not distant". It's confusing.

Instead, explain what is being counted:

These recent measurements are in a short time range. Here is the exclusion count for those measurements.
```suggestion
    'recent_measurement_in_short_time_range_exclusion_count',
```

sbws/lib/v3bwfile.py
```diff
@@ -43,6 +43,25 @@
...
+    # The relay measurements are not X time recent.
+    # By default this is 5 days, which is the same time the older
+    # the measurements can be by default.
+    'recent_measurement_exclusion_not_recent_count',
```
teor2345:
```suggestion
    'recent_measurement_exclusion_too_old_count',
```

sbws/lib/v3bwfile.py
```diff
@@ -43,6 +43,25 @@
...
+    # The minimum number of measurement a relay needs to have.
+    # By default this is two.
+    'recent_measurement_exclusion_not_min_num_count',
```
teor2345:
```suggestion
    'recent_measurement_exclusion_under_min_num_count',
```

sbws/lib/v3bwfile.py
```diff
@@ -135,12 +145,21 @@
     'relay_recent_measurement_failure_count',
     # The number of success results should be:
     # the number of attempts - the number of failures
+    # Number of error results, it can be deduced counting the number of errors
+    # but this way seem clearer.
+    'relay_recent_measurement_exclusion_not_success_count',
```
teor2345:
```suggestion
    'relay_recent_measurement_exclusion_count',
```

sbws/lib/v3bwfile.py
```diff
@@ -135,12 +145,21 @@
...
     # 4.8 relay:  the number of successful results, created in the last 5 days,
-    # that were excluded by a rule, for this relay
+    # that were excluded by a rule, for this relay.
     # This would be the sum of the following 3 + the number of not success
-    'relay_recent_measurement_exclusion_count',
-    'relay_recent_measurement_exclusion_not_distanciated',
+    # 'relay_recent_measurement_exclusion_count',
+    # The number of measurements excluded because they were not X time
+    # away from each other. By default it is one day.
+    'relay_recent_measurement_exclusion_not_distanciated_count',
```
teor2345:
```suggestion
    'relay_recent_measurement_exclusion_short_time_range_count',
```

sbws/lib/v3bwfile.py
```diff
@@ -135,12 +145,21 @@
...
+    # The number of measurements excluded because there were not X time recent.
+    # By default this is 5 days.
     'relay_recent_measurement_exclusion_not_recent_count',
```
teor2345:
```suggestion
    'relay_recent_measurement_exclusion_old_count',
```

sbws/lib/v3bwfile.py
```diff
@@ -135,12 +145,21 @@
...
+    # The number of measurements excluded because there were not at least X.
+    # By default it is 2.
     'relay_recent_measurement_exclusion_not_min_num_count',
```
teor2345:
```suggestion
    'relay_recent_measurement_exclusion_under_min_num_count',
```

sbws/lib/v3bwfile.py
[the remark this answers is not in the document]
juga0: what about `recent_measurement_exclusion_near_count`?

sbws/lib/v3bwfile.py
juga0: what about `recent_measurement_exclusion_few_count`?

juga0:
Also, i would replace all the keys:

`s/_measurement_exclusion_/_measurements_excluded/`

What do you think?

sbws/lib/v3bwfile.py
juga0: I left commented that key because that would be the sum of all the excluded measurements.
For instance, if each rule would exclude a measurement.
What about `relay_recent_measurement_exclusion_error_count`?

sbws/lib/v3bwfile.py
juga0: I left commented that key because that would be the sum of all the relays excluded by a rule.
What about `recent_measurement_exclusion_error_count`?

juga0:
I think i'm also going to modify the code so that the `relay_recent_measurement_exclusion`(*) is not present when it actually does not exclude any measurement.
I think it make it clearer and less data in the file.

sbws/lib/v3bwfile.py
juga0: This is actually the sum of all the errors, which we can count externally, so maybe just not include it?

sbws/lib/v3bwfile.py
juga0: This is actually the sum of all the errors, which we can count externally, so maybe just not include it?

juga0:
I made a "wip" commit with the names i was suggesting, because i don't know if you would like them.

> I think i'm also going to modify the code so that the `relay_recent_measurement_exclusion`(*) is not present when it actually does not exclude any measurement.

In the end i only include them when a measurement is excluded, no matter if the relay is excluded or not.
That commit would need to be squashed if you think is a good solution.
I also change the name of a key in that commit (should have been in the wip commit)

sbws/lib/v3bwfile.py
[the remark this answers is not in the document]
teor2345: Ok

sbws/lib/v3bwfile.py
teor2345: I think we do not need it. Most parsers will be able to add numbers.

sbws/lib/v3bwfile.py
teor2345: Ok

sbws/lib/v3bwfile.py
teor2345: Ok

teor2345:
> Also, i would replace all the keys:
> 
> `s/_measurement_exclusion_/_measurements_excluded/`
> 
> What do you think?

Yes, that seems easier to understand.

teor2345:
> I think i'm also going to modify the code so that the `relay_recent_measurement_exclusion`(*) is not present when it actually does not exclude any measurement.
> I think it make it clearer and less data in the file.

Ok

teor2345:
I will review the new code in about 12-18 hours.

sbws/lib/v3bwfile.py
juga0: hmm, thinking it better, we could now deduce all the relays excluded by a rule counting "vote=0" when it's implemented and then looking why there were measurements excluded for each relay.
But if we leave all the reasons whey the relays are excluded, it'll give us a fast idea on what it's failing.
And then i think we should leave this one for consistency.

sbws/lib/v3bwfile.py
juga0: or maybe leaving it, it makes it more clear that failure (when the relay was attempted to be measured but we don't know what happened) is different from error

juga0:
Added more tests in last fixup

sbws/lib/v3bwfile.py
```diff
@@ -43,13 +43,17 @@
 STATS_KEYVALUES = ['number_eligible_relays', 'minimum_number_eligible_relays',
                    'number_consensus_relays', 'percent_eligible_relays',
                    'minimum_percent_eligible_relays']
+# KeyValues to count the number of relays not included in the bandwidth file
+# of for which is not reported a bandwidth value.
```
teor2345: This comment is really hard to understand, and I don't think it matches the latest code.

Did you mean:
```
# KeyValues that count the number of relays that are in the bandwidth file,
# but ignored by Tor when voting, because they do not have a measured bandwidth.
```

sbws/lib/v3bwfile.py
```diff
@@ -135,11 +138,14 @@
     'relay_recent_measurement_failure_count',
     # The number of success results should be:
     # the number of attempts - the number of failures
```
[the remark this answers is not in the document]
teor2345: I think you missed this fix?

teor2345:
> > Also, i would replace all the keys:
> > `s/_measurement_exclusion_/_measurements_excluded/`
> > What do you think?
> 
> Yes, that seems easier to understand.

I think you missed this change.

sbws/lib/v3bwfile.py
```diff
@@ -746,12 +808,30 @@ def from_results(cls, results, scanner_country=None,
...
         state = State(state_fpath)
+
+        # Create a dictionary with the number of relays excluded by any of the
+        # of the filtering rules that makes relays non-`eligible`.
+        # NOTE: In BW_HEADER_KEYVALUES_RECENT_MEASUREMENTS_EXCLUDED it is
+        # explained what are the KeyValues.
+        # The rules were introduced in #28061 and #27338.
+        # In #28565 we introduce the KeyValues to know why they're excluded.
+        # In #28563 we report these relays, but make Tor ignore them.
+        # This might confirm #28042.
```
teor2345: I don't think you need to repeat this comment?
Repeated comments can get out of sync.
Just say "see from_results() for details"

juga0:
made other fixup, squashed all, force pushed

PR #349, opened 2019-03-13
Ticket28652 rebased

juga0:


sbws/core/scanner.py
```diff
@@ -478,6 +479,14 @@ def main_loop(args, conf, controller, relay_list, circuit_builder, result_dump,
     measured.
 
     """
+    # Variable to count total progress in the last days:
```
asn-d6: Since these are only used for the heartbeat message, how about you track these in the hearbeat module?'

And then instead of doing `[consensus_fp_set.add(r) for r in relay_list.relays_fingerprints]` below, you can add a function to the heartbeat module to register these fingerprints with it (and also clean them if/when they need to be cleaned).

sbws/core/scanner.py
```diff
@@ -503,13 +514,21 @@ def main_loop(args, conf, controller, relay_list, circuit_builder, result_dump,
...
             pending_results.append(async_result)
-
+            measured_fp_set.add(async_result)
         # After the for has finished, the pool has queued all the relays
         # and pending_results has the list of all the AsyncResults.
         # It could also be obtained with pool._cache, which contains
         # a dictionary with AsyncResults as items.
         num_relays_to_measure = len(pending_results)
         wait_for_results(num_relays_to_measure, pending_results)
+        # NOTE: in a future refactor make State a singleton in __init__.py
```
asn-d6: These three lines can go in the heartbeat module as well, right?

sbws/core/scanner.py
juga0: Hmm, the idea there was to register all the relays for which we keep measurements, not just the last consensus.
Unless we make a singelton too to keep track of this, i thinks this variable needs to be left there.
The "old" relays fingerprints are already cleaned in `RelayList`.
Maybe we should not care about previous consensuses and calculate the progress with respect the last consensus? (which seems would make this simpler).

sbws/core/scanner.py
juga0: Yeah, this three can go in the new module

juga0:
The CI failed cause i was missing one "." in the import.

PR #350, opened 2019-03-13
Ticket28983 rebased

juga0:


sbws/lib/v3bwfile.py
```diff
@@ -432,6 +435,42 @@ def add_stats(self, **kwargs):
         [setattr(self, k, str(v)) for k, v in kwargs.items()
          if k in STATS_KEYVALUES]
 
+    def add_relays_excluded_counters(self, exclusion_dict):
+        log.debug("Adding relays excluded counters.")
+        for k, v in exclusion_dict.items():
+            setattr(self, k, str(v))
+
+    def add_time_report_all_network(self):
```
asn-d6: Hm, function name seems misleading. Isn't this supposed to report time for half the network? (Which is also a bit confusing. Perhaps there should be some doc explaining why we are only reporting half of the network and not all of it. I get it but only because I read the ticket).

sbws/lib/v3bwfile.py
```diff
@@ -432,6 +435,42 @@ def add_stats(self, **kwargs):
...
+        number of relays included in the bandwidth file.
+        """
+        # NOTE: in future refactor do not convert attributes to str until
+        # writing to the file.
+        elapsed_time = round(
+            (isostr_to_dt_obj(self.latest_bandwidth)
+             - isostr_to_dt_obj(self.earliest_bandwidth))
+            .total_seconds())
+
+        eligible_relays = getattr(self, 'number_eligible_relay', None)
```
asn-d6: Hm, this seems to check for `number_eligible_relay` which does not exist? Is this meant to check for `number_eligible_relays`?

Does getattr work with substrings, is this broken, or something else that I don't get?
Do we need tests for this code?

In general, I'm not a fan of programming with `getattr` because it's the type of meta-programming that is really prone to errors. If it was me, I would initialize all those fields to None on the constructor, and then check that they are not None here.

sbws/lib/v3bwfile.py
```diff
@@ -432,6 +435,42 @@ def add_stats(self, **kwargs):
         [setattr(self, k, str(v)) for k, v in kwargs.items()
          if k in STATS_KEYVALUES]
 
+    def add_relays_excluded_counters(self, exclusion_dict):
```
asn-d6: Hm, who is calling this func?

sbws/lib/v3bwfile.py
juga0: yeah, should be call `half_network`
ok, will add doc to explain why half

sbws/lib/v3bwfile.py
juga0: hmm, i think this might be coming from other ticket/branch and added here by mistake, will check

sbws/lib/v3bwfile.py
juga0: > Hm, this seems to check for number_eligible_relay which does not exist? Is this meant to check for number_eligible_relays?

yes
> Does getattr work with substrings, is this broken, or something else that I don't get?
> Do we need tests for this code?
Test would be good, but i t would prefer to do delay it after some refactoring #28684

> In general, I'm not a fan of programming with getattr because it's the type of meta-programming that is really prone to errors. If it was me, I would initialize all those fields to None on the constructor, and then check that they are not None here.

I understand your reasons.
I started to do this because we're adding new attributes progressively and when parsing old files this would make it easier to just ignore the attributes that were not present.
I think the right way to do it though would be as steam does: https://github.com/torproject/stem/blob/master/stem/descriptor/bandwidth_file.py#L61

sbws/lib/v3bwfile.py
juga0: yes, it was added in https://github.com/torproject/sbws/pull/348/commits/5fd38798a8070818d0b521b56b5ca50943af4ca7#diff-0c88fa36ad3aabea4aa3c57e63f54db8R441
(i should work on less tickets at a time).

sbws/lib/v3bwfile.py
juga0: > Tests would be good, but i t would prefer to do delay it after some refactoring #28684
ok, i think i'll add a test

juga0:
Summary of changes in the fixup:
I added the test
I do not round the time and instead round for the final values
I check first that the needed attributes are present
Calculate the actual half-life (before it was wrong)
Replace the percent for the actual number
Replace log.debug by info
Replace minutes by hours

tests/unit/lib/test_v3bwfile.py
```diff
@@ -393,3 +394,17 @@ def test_update_progress(datadir, tmpdir):
...
+def test_time_measure_half_network(caplog):
+    header = V3BWHeader(timestamp_l,
+                        file_created=file_created,
+                        generator_started=generator_started,
+                        earliest_bandwidth=earliest_bandwidth)
+    header.number_consensus_relays = '6500'
+    header.number_eligible_relays = '4000'
+    caplog.set_level(logging.INFO)
+    header.add_time_report_half_network()
+    assert header.time_to_report_half_network == '70200'  # 20h
```
asn-d6: Hm. Shouldn't the full time of "20h" be 40 hours? Why are we testing for 39 hours?
Is the comment wrong, the logic weird, or something else?

juga0: ha, yeah, sorry, i rounded 19.5 and didn't commented. Fixup with no rounding.

juga0:
squashed and merged

PR #351, opened 2019-03-14
chg: docs: document that authorities are not measured

juga0:
and improve a bit the summary about what sbws does.
Closes: #29722


juga0:
squashed and merged

PR #352, opened 2019-03-16
Ticket29589

juga0:


sbws/lib/destination.py
```diff
@@ -91,80 +99,166 @@ def connect_to_destination_over_circuit(dest, circ_id, session, cont, max_dl):
...
+                 self.url, self._delta_seconds_retry / 60 / 60)
+
+    def _is_last_try_old_enough(self, n=None):
+        """
+        Return True if the last time it was used it was ``n`` seconds ago.
+        """
+        # Timestamp of the last attempt.
+        last_time = self._attempts[-1][0]
+        # If the last attempt is older than _delta_seconds_retry,
+        if (datetime.datetime.utcnow()
```
ahf: This can be shortened to `return (datetime.datetime.utcnow() - datetime.timedelta(seconds=self._delta_seconds_retry)) > last_time`

juga0:
squashed and merged

PR #353, opened 2019-03-18
Ticket28563 rebased ticket28565

juga0:


sbws/lib/v3bwfile.py
```diff
@@ -544,29 +550,56 @@ def from_results(cls, results, secs_recent=None, secs_away=None,
         # In #28563 will include again these lines, but make them unparseable
         # by Tor.
         # This might confirm #28042.
+        exclusion_reason = None
         kwargs['relay_recent_measurement_exclusion_not_success_count'] = \
             len(results) - len(success_results)
         if not success_results:
-            return None, 'recent_measurement_exclusion_not_success_count'
+            exclusion_reason = 'recent_measurement_exclusion_not_success_count'
```
nmathewson: I'm concerned by the change in logic by moving all of these "return" statements.  The way that the code was written before, it would return the _first_ reason that it found for excluding the relay.  But now, it looks like it will return the _last_ reason.  Do we really want to make that change here?

sbws/lib/v3bwfile.py
```diff
@@ -544,29 +550,56 @@ def from_results(cls, results, secs_recent=None, secs_away=None,
...
+        # NOTE: the bandwidth values could still be obtained if:
+        # 1. ``ResultError`` will store them
+        # 2. assign ``results_recent = results`` when there is a ``exclusion
+        # reason.
+        # This could be done in a better way as part of a refactor #28684.
+        if exclusion_reason is not None:
+            return (cls(node_id, 1, **kwargs), exclusion_reason)
+
+        # For any line not excluded, change vote and unmeasured values
+        # Maybe we should just not include these KeyValues in this case.
```
nmathewson: I think that makes more sense.

sbws/lib/v3bwfile.py
```diff
@@ -544,29 +550,56 @@ def from_results(cls, results, secs_recent=None, secs_away=None,
...
+            exclusion_reason = \
+                'recent_measurement_exclusion_not_min_num_count'
+
+        # Create a bandwidth line with the relay, but set ``vote=0`` so that
+        # Tor versions with patch #29806 does not vote on the relay.
+        # Set ``bw=0`` so that Tor versions without the patch,
+        # will give the relay low bandwidth.
+        # Include ``unmeasured=1`` in case Tor would vote on unmeasured relays
+        # in future versions.
+        kwargs['vote'] = '0'
```
nmathewson: It might make more sense to only set these if we have an exclusion_reason.

sbws/lib/v3bwfile.py
```diff
@@ -1201,27 +1242,27 @@ def is_min_perc(self):
 
     @property
     def sum_bw(self):
-        return sum([l.bw for l in self.bw_lines])
+        return sum([l.bw for l in self.bw_lines if getattr(l, 'bw', None)])
```
nmathewson: Maybe use hasattr in these checks here instead of getattr?  I think it is closer to what you mean.

sbws/lib/v3bwfile.py
```diff
@@ -1201,27 +1242,27 @@ def is_min_perc(self):
...
 
     @property
     def num(self):
         return len(self.bw_lines)
 
     @property
     def mean_bw(self):
-        return mean([l.bw for l in self.bw_lines])
+        return mean([l.bw for l in self.bw_lines if getattr(l, 'bw', None)])
```
nmathewson: This one needs to be hasattr rather than getattr, I think, or else the mean won't include any nodes with bw=0.

sbws/lib/v3bwfile.py
juga0: You are totally right, i left it to return on each condition as it was, just return also the reason

sbws/lib/v3bwfile.py
juga0: Changed

sbws/lib/v3bwfile.py
juga0: Changed

sbws/lib/v3bwfile.py
juga0: i forgot hasattr exists, yeah, changed, also all the other lines, because i don't want to take them into account if they don't have the attribute.

juga0:
I also added tests to check vote and unmeasured

juga0:
squashed and merged

PR #356, opened 2019-03-25
Bug29891

juga0:


sbws/core/scanner.py
```diff
@@ -98,8 +97,11 @@ def timed_recv_from_server(session, dest, byte_range):
     try:
         # headers are merged with the session ones, not overwritten.
         session.get(dest.url, headers=HTTP_GET_HEADERS, verify=dest.verify)
-    # Catch any `requests` exception, so that it can stored in the Result
-    except requests.exceptions.RequestException as e:
+    # All `requests` exceptions could be catch with
```
ahf: I'd use 'caught' here.

sbws/core/scanner.py
```diff
@@ -261,20 +263,25 @@ def measure_relay(args, conf, destinations, cb, rl, relay):
                 ]
     # Pick a destionation
     dest = destinations.next()
-    # If there is no any destination at this point, it can not continue.
+    # When there're no any functional destinations.
     if not dest:
-        # XXX: this should return a ResultError
-        # instead of stopping the scanner once a destination can be recovered.
+        # NOTE: When there're still functional destinations but only one of
```
ahf: I think there is missing some binding word? "When there's still an available functional destination, but ..." ?

sbws/core/scanner.py
```diff
@@ -522,7 +542,10 @@ def main_loop(args, conf, controller, relay_list, circuit_builder, result_dump,
 
         loop_tstop = time.time()
         loop_tdelta = (loop_tstop - loop_tstart) / 60
-        log.debug("Measured %s relays in %s minutes", num_relays, loop_tdelta)
+        # At this point, we know the relays that were queued to be measured,
+        # what does not mean they were actually measured.
```
ahf: Should 'what' be changed to 'that'?

PR #357, opened 2019-05-06
Ticket30406 refactor header constants

juga0:


sbws/lib/stem_bandwidth_file.py
```diff
@@ -0,0 +1,67 @@
+"""Stem's Bandwidth file parser part.
```
asn-d6: Hmm, is this the right way to do this? Just include a file from stem? Doesn't this mean we need to keep track of stem's revisions to this file? :/

Also if this is the right way to do this, then perhaps we should mention which revision of this file this is (commit id). Also, why does it say "To be removed on Stem's release 1.8.0"?

sbws/lib/v3bwfile.py
```diff
@@ -44,74 +46,82 @@
...
-    + BW_HEADER_KEYVALUES_MONITOR
-
-KEYVALUES_INT = STATS_KEYVALUES + BW_HEADER_KEYVALUES_MONITOR
-# List of all unordered KeyValues currently being used to generate the file
-UNORDERED_KEYVALUES = EXTRA_ARG_KEYVALUES + STATS_KEYVALUES + \
-                      ['latest_bandwidth'] + \
-                      BW_HEADER_KEYVALUES_MONITOR
-# List of all the KeyValues currently being used to generate the file
-ALL_KEYVALUES = ['version'] + UNORDERED_KEYVALUES
+# # KeyValues that count the number of relays that are in the bandwidth file,
```
asn-d6: Why do we keep all this commented code around?

sbws/lib/v3bwfile.py
```diff
@@ -250,20 +260,28 @@ class V3BWHeader(object):
...
         assert isinstance(timestamp, str)
         for v in kwargs.values():
             assert isinstance(v, str)
         self.timestamp = timestamp
-        # KeyValues with default value when not given by kwargs
+
+        # Initialize all defined attrs,
+        # XXX: Since using the stem's name for the attributes will require
+        # other refactoring here, use the name in the actual file.
+        header_keys = [k for k, _ in HEADER_ATTR.values()]
```
asn-d6: It's not easy to understand if this code is refactoring or changes the behavior of sbws (adds/removes headers). 

Could we have a unittest which makes sure that the old code and the new code will output the exact same thing? Perhaps instead of commenting out the code above, you can move it to the tests directory and make a unittest that checks the output of the old with the new code. Would this make sense?

sbws/lib/stem_bandwidth_file.py
juga0: > Hmm, is this the right way to do this? Just include a file from stem? Doesn't this mean we need to keep track of stem's revisions to this file? :/

Only until the next release, i think there're stem's releases every ~6 months.
And so far this file has only changed when after adding more headers to `sbws` i create an issue in `stem` so it gets added too.
With this change, instead i could directly create a PR in stem and duplicate it in `sbws` until the next release.

Not sure is there's a better way, but i definitely don't want to wait 8 months again to do a sbws release waiting for a stem release and it's not a solution to change the dependency to stem's git master because that does not work for system packages (Debian).

> 
> Also if this is the right way to do this, then perhaps we should mention which revision of this file this is (commit id).

Do you think it would be better i just add the stem's bandwidth_file.py as it is currently?.
I can add the commit id (in v3bwfile if i add the stem's bandwidth_file.py as it's)

> Also, why does it say "To be removed on Stem's release 1.8.0"?

Because that's the next Stem's release, i can add this comment somewhere. Where it would fit better?.

sbws/lib/v3bwfile.py
juga0: > Why do we keep all this commented code around?

https://github.com/torproject/sbws/pull/357/files/8a5efb16a8bf25656d1fb515ac43202542f1dcec..ebbb9ffcf3ee28d48c68b7ef45463ed5f534e2ff#diff-0c88fa36ad3aabea4aa3c57e63f54db8R76:

```
# The commented code is kept so that
# it can be known what each Key means.
```

But certainly, i can just link to the specification where they are explained. And anyway those comments should now be in stem. Sounds good?

sbws/lib/v3bwfile.py
juga0: > It's not easy to understand if this code is refactoring or changes the behavior of sbws (adds/removes headers).
> 
> Could we have a unittest which makes sure that the old code and the new code will output the exact same thing? Perhaps instead of commenting out the code above, you can move it to the tests directory and make a unittest that checks the output of the old with the new code. Would this make sense?

Makes sense but then need to keep this old code in the test.
Stem has already tests for all the bandwidth files versions using data generated from sbws.
I checked, and including the stem's tests would require adding many more files.

So i came out with other solution that it's something i wanted to do time ago: include real measurements and bandwidth files generated with the test network for the tests.

So i run the integration tests to generate those files with master branch, then i created a test that generates the bandwidth file from the previous results and compare that the bandwidth file is the same.

I'm not pushing these changes yet until knowing your opinion. If you don't like that solution will open other ticket to include them.

juga0: Other way of ensuring that i didn't accidentally change functionality is the fact that i didn't have to change any test for this PR. It's true though that i don't think there're unit tests checking all the headers, but the previous idea would solve that.

sbws/lib/stem_bandwidth_file.py
juga0: if i add the file as it is, i also need to copy the `__init__.py`
Thinking more about this, i think it's better to leave only that part, to follow better what i substitute.
I could also do a `try import` and if there's no `bandwidth_file.py` in stem then leave the old code + the refactor.
But i think duplicating all that code when as i said, the file won't change unless someone open a ticket, and there'll be a new stem release soon.

juga0: See small changes in fixup

sbws/lib/v3bwfile.py
juga0: Removed commented code

sbws/lib/v3bwfile.py
juga0: Pushed the new test

sbws/lib/v3bwfile.py
asn-d6: ACK!

sbws/lib/v3bwfile.py
```diff
@@ -15,6 +19,11 @@
                           TORFLOW_SCALING, SBWS_SCALING, TORFLOW_BW_MARGIN,
                           TORFLOW_OBS_LAST, TORFLOW_OBS_MEAN,
                           PROP276_ROUND_DIG, MIN_REPORT, MAX_BW_DIFF_PERC)
+
+# XXX: replace this line by:
+# from stem.descriptor.bandwidth_file import HEADER_ATTR, _int
+# when stem releases 1.8.0. More information in stem_bandwidth_file.py
+from .stem_bandwidth_file import HEADER_ATTR, _int
```
teor2345: Please write CI tests for sbws that use stem 1.7.0 and stem master.

tests/data/.sbws.ini
```diff
@@ -0,0 +1,30 @@
+[general]
+# Days into the past that measurements are considered valid
+data_period = 1
+
+[paths]
+sbws_home = .sbws
```
asn-d6: Perhaps we don't need to hide this data in .sbws? I'm not used to files in source code being hidden. Why are we doing it here?

tests/data/.sbws/tor/cached-consensus
```diff
@@ -1,9 +1,9 @@
 network-status-version 3
 vote-status consensus
 consensus-method 28
-valid-after 2019-03-25 13:08:50
-fresh-until 2019-03-25 13:09:00
-valid-until 2019-03-25 13:09:20
+valid-after 2019-05-11 06:32:10
```
asn-d6: Why did we need to update the consensus?

sbws/lib/v3bwfile.py
asn-d6: I like this new test very much. I think it's a good step.
I added a few comments while I was trying to understand what the test does.

PR #358, opened 2019-06-16
fix: globals: Fetch descriptors early

juga0:
So that sbws detect early changes in the relay descriptors.

Closes: #30733. Bugfix v0.2.0.

### DIFF
--- a/sbws/globals.py
+++ b/sbws/globals.py
@@ -36,6 +36,28 @@ TORRC_STARTING_POINT = {
     'SafeLogging': '0',
     'LogTimeGranularity': '1',
     'ProtocolWarnings': '1',
+    # Text and options from
+    # https://stem.torproject.org/tutorials/mirror_mirror_on_the_wall.html#can-i-get-descriptors-from-the-tor-process
+    #
+    # Descriptors have a range of time during which they're valid. To get the
+    # most recent descriptor information, regardless of if Tor needs it or not,
+    # set the following.
+    'FetchDirInfoEarly':  '1',
+    'FetchDirInfoExtraEarly': '1',
+
+    # Tor doesn't need all descriptors to function. In particular...
+    #
+    #   * Tor no longer downloads server descriptors by default, opting
+    #     for microdescriptors instead.
+    #
+    #   * If you aren't actively using Tor as a client then Tor will
+    #     eventually stop downloading descriptor information altogether
+    #     to relieve load on the network.
+    #
+    # To download descriptors regardless of if they're needed by the
+    # Tor process or not set...
+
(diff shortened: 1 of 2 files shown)


sbws/globals.py
```diff
@@ -36,6 +36,28 @@
...
+    # Descriptors have a range of time during which they're valid. To get the
+    # most recent descriptor information, regardless of if Tor needs it or not,
+    # set the following.
+    'FetchDirInfoEarly':  '1',
+    'FetchDirInfoExtraEarly': '1',
+
+    # Tor doesn't need all descriptors to function. In particular...
+    #
+    #   * Tor no longer downloads server descriptors by default, opting
+    #     for microdescriptors instead.
```
teor2345: This comment is wrong, because we set UseMicrodescriptors 0 above.

sbws/globals.py
```diff
@@ -36,6 +36,28 @@
...
+    #
+    #   * If you aren't actively using Tor as a client then Tor will
+    #     eventually stop downloading descriptor information altogether
+    #     to relieve load on the network.
```
teor2345: This is a good reason for us to use FetchUselessDescriptors: we don't want Tor going idle.

sbws/globals.py
juga0: I guess it makes sense to have it in the context of stem, not sbws.
Fixup removed the context.

juga0: s/context/comment/

juga0:
> This comment is also wrong:
> 
> "To download descriptors regardless of if they're needed by the
> Tor process or not set..."

Replaced with actual man comment on the opiton.

teor2345:
> This code contains a race condition: new_relays_dict is modified while its keys are being iterated.
> Python 3.8 will warn when this happens, maybe older versions do too.

I opened https://trac.torproject.org/projects/tor/ticket/30908 for this issue

juga0:
> This code contains a race condition: new_relays_dict is modified while its keys are being iterated.
> Python 3.8 will warn when this happens, maybe older versions do too.

Hmm, i'm not iterating new_relays_dict but relays. Are you sure about this?, do you have a log with python 3.8?.

> To avoid the race condition, take a copy of the keys:
> if r.fingerprint in list(new_relays_dict.keys()):

That's a condition, not a loop.
 
> See this stem commit for a similar bug and fix:
> https://gitweb.torproject.org/stem.git/commit/?id=b5aecb743d33db1a6378d59792d8e57305b6c6f2

It seems to me a different case

teor2345:
> It seems to me a different case

You're right, I closed the ticket as "not a bug".

juga0:
Closing in favor of #362

PR #360, opened 2019-06-17
relaylist: stop using the current time when a consensus is downloaded…

teor2345:
… twice

Instead:
* use the consensus valid-after time, or
* use the supplied timestamp, or
* warn and use the current time.

This should fix the occasional CI failure, when the current time is 1 second
later than the test consensus time. (Or it should warn, and we can fix the
test code.)

Fixes bug 30909; bugfix on 1.1.0.

### DIFF
--- a/sbws/lib/relaylist.py
+++ b/sbws/lib/relaylist.py
@@ -175,34 +175,47 @@ class Relay:
             return self._consensus_timestamps[-1]
         return None
 
+    def _append_consensus_timestamp_if_later(self, timestamp):
+        """Append timestamp to the list of consensus timestamps, if it is later
+           than the most recent existing timestamp, or there are no timestamps.
+           Should only be called by _add_consensus_timestamp().
+           timestamp must not be None, and it must not be zero.
+        """
+        if not timestamp:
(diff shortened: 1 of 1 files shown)


juga0:
lgtm, just a comment.
I think this is a case where you would have prefer a commit with the change and a commit refactoring the code to a function. Wouldn't you?.

teor2345:
I don't know how to do the refactor without fixing the bug.

juga0:
> I don't know how to do the refactor without fixing the bug.

Repeating the code where it is call. For instance: https://github.com/juga0/sbws/commit/5ad8f372a2bd735119abeb5cb54caa493385c0ba

juga0:
I realized there's another bug, when `_consensus_timestamps` is None, which should not be because of `__init__`. I based a branch on yours and fixed that in https://github.com/juga0/sbws/commit/26ecec60ab464725db16d4e29e78bd518028b156, i also added tests.

juga0:
Another way to solve the timestamp used as the consensus timestamp is to refactor the code to first decide on which timestamp to use, as in https://github.com/juga0/sbws/commit/76c7f8d86a42f9bbc36140e300a8632868036119

PR #361, opened 2020-01-26
Maint 1.1 issue30727 measure known relays

juga0:


sbws/lib/relaylist.py
```diff
@@ -288,6 +289,12 @@ def increment_relay_recent_priority_list_count(self):
             self.relay_recent_priority_list_count = 0
         self.relay_recent_priority_list_count += 1
 
+    def is_old(self):
+        """Whether the last consensus seen for this relay is older than the
+        measurement period.
+        """
+        timestamp.is_old(self.last_consensus_timestamp)
```
ahf: Shouldn't this "return" the value from `is_old()` ?

sbws/lib/relaylist.py
```diff
@@ -415,28 +422,49 @@ def _init_relays(self):
         # Change to stem.descriptor.remote in future refactor.
         network_statuses = c.get_network_statuses()
         new_relays_dict = dict([(r.fingerprint, r) for r in network_statuses])
+        log.debug("Number of relays in the current consensus: %s.",
+                  len(new_relays_dict))
```
ahf: `%d` instead of `%s`?

sbws/lib/relaylist.py
```diff
@@ -415,28 +422,49 @@ def _init_relays(self):
...
+
         new_relays = []
+
+        # Only or debugging, count the relays that are not in the current
+        # consensus and have not been seen in the last consensuses either.
+        num_old_relays = 0
+
         relays = copy.deepcopy(self._relays)
         for r in relays:
             if r.fingerprint in new_relays_dict.keys():
```
ahf: I know this change is not for this patch, but one can do `if key in dict` without first generating a list of keys and check for an item in that list.

sbws/util/timestamp.py
```diff
@@ -71,3 +73,22 @@ def unixts_to_str(unixts):
     # than int or float
     assert isinstance(unixts, int) or isinstance(unixts, float)
     return str(unixts)
+
+
+# XXX: tech-debt: replace all the code that check whether a
+# measurement or relay is older than the measurement period by this.
+def is_old(timestamp, measurements_period=MEASUREMENTS_PERIOD):
```
ahf: Good!

sbws/lib/relaylist.py
juga0: auch, totally

sbws/lib/relaylist.py
juga0: hmm, it's python, i think it doesn't really matter. I changed, then changed the rest of logging lines where i'm using a number

sbws/lib/relaylist.py
juga0: hmm, we're generating 2 new things here: the dictionary to search for keys and an a copy of the list of relays. While the later isn't really needed and could just be modified, the former is useful to be able to easily search in a dictionary.
Ah, or you mean that it could be `if r.fingerprint in new_relays_dict` without the last `.keys()`

teor2345: It's much faster if you don't use .keys().

Searching a list takes a time that depends on the length of the list (and the position of the relay in the list). That could be a long time for a big list of relays.

Searching a dict usually takes a shorter amount of time, because it's a hash table.

PR #367, opened 2020-02-09
Maint 1.1 issue30196 01 tor version

juga0:


sbws/lib/v3bwfile.py
```diff
@@ -32,22 +32,44 @@
...
+]
+
 # number_eligible_relays is the number that ends in the bandwidth file
 # ie, have not been excluded by one of the filters in 4. below
 # They should be call recent_measurement_included_count to be congruent
 # with the other KeyValues.
-STATS_KEYVALUES = ['number_eligible_relays', 'minimum_number_eligible_relays',
-                   'number_consensus_relays', 'percent_eligible_relays',
-                   'minimum_percent_eligible_relays']
+HEADER_KEYS_V1_2 = [
```
teor2345: Please put these arrays in version order: V1_2 before V1_3

sbws/lib/v3bwfile.py
```diff
@@ -136,13 +175,19 @@
...
+BWLINE_KEYS_V1_2 = [
+    "bw_median",
+    "bw_mean",
+    "desc_bw_avg",
+    "desc_bw_bur",
+    "desc_bw_obs_last",
+    "desc_bw_obs_mean",
+    "consensus_bandwidth",
+    "consensus_bandwidth_is_unmeasured",
+]
```
teor2345: Please add a comment that explains that there were no BWLINE keys added in V1_3

sbws/lib/v3bwfile.py
```diff
@@ -23,7 +23,7 @@
 
 log = logging.getLogger(__name__)
 
-LINE_SEP = '\n'
+LINES_SEP = '\n'
```
teor2345: This rename seems unnecessary?

sbws/lib/v3bwfile.py
```diff
@@ -32,22 +32,44 @@
...
 # =================
-# List of the extra KeyValues accepted by the class
-EXTRA_ARG_KEYVALUES = ['software', 'software_version', 'file_created',
-                       'earliest_bandwidth', 'generator_started',
-                       'scanner_country', 'destinations_countries']
+# KeyValues that need to be in a specific order in the Bandwidth File.
+HEADER_KEYS_V1_1_ORDERED = ['version']
+# KeyValues that are not initialized from arguments nor depend on the state
+# file nor the measurements.
+HEADER_KEYS_V1_1_SELF_INIT = [
```
teor2345: The difference between SELF_INIT and TO_INIT is unclear, please rename.
I suggest "ARG" and "EXTERN". (Arguments or something external.)

I also wonder if latest_bandwidth does depend on the state file?
If there are new bandwidths (or no bandwidths at all), then the latest bandwidth should change.

sbws/lib/v3bwfile.py
```diff
@@ -32,22 +32,44 @@
...
+    "latest_bandwidth",
+]
+# KeyValues that are initialized from arguments.
+HEADER_KEYS_V1_1_TO_INIT = [
+    "earliest_bandwidth",
+    "generator_started"
+]
+
+# KeyValues added in the Bandwidth File v1.3.0
+HEADER_KEYS_V1_3 = [
```
teor2345: Maybe the header keys for the other versions should also end in ARG?

teor2345: Or EXTERN? Where do they actually come from?

sbws/lib/v3bwfile.py
```diff
@@ -91,34 +113,51 @@
...
+# KeyValues that are initialized from arguments, not self-initialized.
+HEADER_INIT_KEYS = (
+    HEADER_KEYS_V1_1_TO_INIT
+    + HEADER_KEYS_V1_3
+    + HEADER_KEYS_V1_2
+    + HEADER_KEYS_V1_4
+    + HEADER_KEYS_V1_5_TO_INIT
+)
+
+HEADER_INT_KEYS = HEADER_KEYS_V1_2 + HEADER_KEYS_V1_4
```
teor2345: INT is very similar to INIT, the renames above will fix these confusing names.

scripts/code_replacements/rename_constants_bwfile.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
teor2345: Why are you committing this file?

After it's run once to produce this commit, it's not going to be used.
Could you put the script in the commit message instead?

teor2345: Also, you can make the script shorter by doing "find sbws tests …"

juga0:
Merged https://github.com/juga0/sbws/tree/maint-1.1_issue30196_01_tor_version_squashed, there's a `,` difference.

PR #370, opened 2020-02-20
fix: Add at build time the git revion to version

juga0:
Instead of having a hardcoded version, calculate the version at build
time making use of `git describe --tags --dirty --always`.

This way, even if the program is not running from inside a git
repository it still can know which was the git revision from the
source it was installed from.
If the program is launched from a path that is a git repository, it
does not gives the git revision of that other repository.

If's also able to get the version when installed from a tarball.
It does not add the git revision when it's being install from a git
tag.

`versioneer` external program is only needed the first time, because
it copies itself into the repository. So it does not add an external
dependency.

There're no changes needed to the `--version` cli argument nor to the
code that generates the bandwidth file, since they both use the
variable `__version__`.

The version previous to this commit was `1.1.1-dev0`, after
this commit, it becomes `1.1.0+xx.gyyyyyyyy`, ie. xx commits after
`1.1.0` plus the git short hash (yyyyyyyy).

Closes: #30899.


teor2345:
This is a great change!

Do you also want to use --broken, to show if the repository is corrupt?
https://git-scm.com/docs/git-describe

juga0:
> This is a great change!
> 
> Do you also want to use --broken, to show if the repository is corrupt?
> https://git-scm.com/docs/git-describe

i'd not do that cause the code itself it's automatically generated by https://github.com/warner/python-versioneer.
See my comment in Trac.

sbws/__init__.py
```diff
@@ -1,4 +1,6 @@
-__version__ = '1.1.1-dev0'
+from ._version import get_versions
+__version__ = get_versions()['version']
+del get_versions
```
ahf: Is this to ensure the symbol doesn't get (re-)exported when one does `import sbws` ?

sbws/_version.py
```diff
@@ -0,0 +1,520 @@
...
+    """Create, populate and return the VersioneerConfig() object."""
+    # these strings are filled in when 'setup.py versioneer' creates
+    # _version.py
+    cfg = VersioneerConfig()
+    cfg.VCS = "git"
+    cfg.style = "pep440"
+    cfg.tag_prefix = "v"
+    cfg.parentdir_prefix = "sbws-"
+    cfg.versionfile_source = "sbws/_version.py"
+    cfg.verbose = False
```
ahf: Why is this both here and in the config file? "pep440" is defined both places? Maybe they have to be?

juga0: the first time versioneer is run, it check the config and writes to _version.py
config is not checked afterwards.
i think i could be removed from config, but it does not hurt there and maybe is useful at some point to compare the versioneer.py file we have with some other.

sbws/__init__.py
juga0: i think i to don't have that function imported whenever sbws i imported.

juga0:
i added an extra commit explaining why we choose to use versioneer.
i think it can be helpful to document this type of decisions, since i wasn't too sure on using this and other people or us might wonder in the future.

